PR #297, opened 2013-09-04
Contexts schema.org

ivikash:
Auto-generated contexts for schema.org


gkellogg:
Now that schema.org maintains it's own context, this PR is obsolete.

PR #318, opened 2014-02-09
Adding fullscreen of editor area in playground

bollwyvl:
This adds a button to toggle fullscreen mode on the playground without the header and footer and introductory text.

Here is a [demo](http://bollwyvl.github.io/json-ld.org/playground-fullscreen/playground/).

This approach gets around some of the limitations of the outer responsive bootstrap `container` by only fullscreening a second, interior `container-fluid`. This does make the "flash of unstyled textareas" issue more pronounced (expecially on a slow connection), so it could be that needs to be solved first.

### DIFF
--- a/playground/index.html
+++ b/playground/index.html
@@ -112,9 +112,6 @@
         </div>
       
         <div class="pull-right">
-          <button class="btn fullscreen">
-            <i class="icon-fullscreen"></i> Fullscreen
-          </button>
           <button class="btn popover-info">
             <i class="icon-magic"></i> Shortcuts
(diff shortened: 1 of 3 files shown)


brutalcrozt:
ping and :ear:

gkellogg:
This PR is languishing, we should either commit to close.

cc/ @msporny @davidlehn @dlongley @lanthaler.

gkellogg:
@bollwyvl, if you'd care to create a new PR, we can move forward, closing this for now as old and conflicting.

PR #332, opened 2014-03-06
Swapped RFC4627 for RFC7159.

tomayac:


tomayac:
Closing, as this has not progressed since March 6, 2014 and no longer really matters.

PR #378, opened 2015-01-25
adding basic gist integration

bollwyvl:
As per a [recent discussion](https://lists.w3.org/Archives/Public/public-linked-json/2015Jan/0030.html) on the mailing list (cc @msporny), this adds (anonymous) gist (de)serialization to the playground.

Here is an example URL for loading a gist:
http://bollwyvl.github.io/json-ld.org/playground/#/gist/12397f939500d17a1a9f

With the supporting gist:
https://gist.github.com/anonymous/12397f939500d17a1a9f

This is really more of a starting point than anything else.

- There is a bit of a mismatch between what is shown vs what is stored: the name, which is just a good idea for the gist, doesn't have anyplace to live.
- At present, there is no way to create a gist as a particular user, which would be nice... last I checked that required a minimal oauth backend to do properly.
- On the list, mentioned comments, etc. not sure how these would come about

### DIFF
--- a/playground/index.html
+++ b/playground/index.html
@@ -124,6 +124,10 @@
             <i class="icon icon-link"></i>
             <span>Permalink</span>
           </a>
+          <a class="btn" id="gist">
+            <i class="icon icon-github"></i>
+            <span>Gist</span>
+          </a>
           <button class="btn popover-info" title="Keyboard shortcuts">
             <i class="icon icon-keyboard"></i> Shortcuts
           </button>
--- a/playground/playground.js
+++ b/playground/playground.js
@@ -149,6 +149,60 @@
       return rval;
     }
 
+    function handleGist(gist){
+      /**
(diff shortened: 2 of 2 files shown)


msporny:
Awesome, @bollwyvl - agreed that it's a start and we may want to refine how it's implemented a bit, but really happy w/ this addition.

Thank you! You rock!

dlongley:
:+1:

bollwyvl:
Ok! That was quick! Maybe we slap a little beta symbol on it in the meantime :)

I would hate to create _yet another_ backwards-compatibility issue because the thing I hacked together in an afternoon wasn't well thought out. I just kinda fiat-ed some things in a playground namespace.

To that end, I guess I'd like some more feedback on the notional structure of a playground document `playground.jsonld` that can a) handle today's challenge and b) handle more of the multi-format (yaml, etc) thing talked about on the list.

PR #380, opened 2015-01-25
Add Initial Activity Streams 2.0 Example

jasnell:
Adds an initial Activity Streams 2.0 example along with a redirection for the 
development context document.

### DIFF
--- a/playground/index.html
+++ b/playground/index.html
@@ -117,6 +117,10 @@
             <i class="icon icon-book"></i>
             <span>Library</span>
           </button>
+          <button id="btn-activity" class="btn button">
+            <i class="icon icon-comment"></i>
+            <span>Activity</span>
+          </button>
         </div>
 
         <div class="pull-right">
@@ -217,7 +221,7 @@
         <div id="markup-errors" class="text-error"></div>
         <div id="param-errors" class="text-error"></div>
         <div id="using-context-map" class="hide alert alert-note">
-          <p>NOTE: A remote context that is known to be fully working yet was detected in your input.
+          <p>NOTE: A remote context that is not known to be fully working yet was detected in your input.
             If you wish, you can use an alternative context created by the JSON-LD community to
             process your document nevertheless.</p>
           <label class="checkbox">
--- a/playground/playground-examples.js
+++ b/playground/playground-examples.js
@@ -171,4 +171,20 @@
     }
   };
 
+  // add an Activity Streams 2.0 Example
+  playground.examples["Activity"] = {
+    "@context": "http://www.w3.org/ns/activitystreams#",
+    "@type": "Post",
+    "actor": {
+      "@type": "Person",
+      "@id": "acct:sally@example.org",
+      "displayName": "Sally"
+    },
+    "object": {
+      "@type": "Note",
+      "content": "This is a simple note"
+    },
+    "published": "2015-01-25T12:34:56Z"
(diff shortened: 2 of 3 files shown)


msporny:
Thanks for the example! :)

Any chance that you might alias "@id" -> "id", and "@type" -> "type" in the AS2 JSON-LD Context in the future?

msporny:
Also, any chance you could add the CORS headers to the AS2 JSON-LD Context? That "Use development XYZ" is a bit strange for something that you should have control over.

jasnell:
The `http://www.w3.org/ns/activitystreams#` url to host the AS2 JSON-LD context document has not yet been set up. I currently have the context document hosted at `http://asjsonld.mybluemix.net` and it does have the CORS headers set. Because the "official" url is not yet set up we need to have the redirect in there.

And it's not likely that we'll map "@id" to "id" and "@type" to "type". Any particular reason why on those?

jasnell:
I'm working on it ;-) ... I can definitely change the example to point to the current dev context location and update it when the official one is finally set up.

jasnell:
https://github.com/json-ld/json-ld.org/pull/381

msporny:
> re: problem w/ the official URL

ETA on when this'll be fixed?

> not likely that we'll map "@id" to "id" and "@type" to "type". Any particular reason why on those?

Because it makes the data look more foreign than it needs to be to people that are used to JSON, but not used to JSON-LD. Specifically, it's nice to be able to use dot-notation in the languages that support it:
```javascript
if(obj.id === theOneImLookingFor) {
```
instead of this:

```javascript
if(obj['@id'] === theOneImLookingFor) {
```

The top one is easier to parse while looking through code.

jasnell:
No ETA yet. I'll follow up with the W3C contact this next week.

akuckartz:
@msporny That the "@" in front of "id" makes the data look more foreign than without is an interesting opinion. It _might_ be true though. I wonder what you would do if you could rewrite the JSON-LD spec.

My opinion is different: I think the "@" indicates that the data is machine readable in an additional sense.

msporny:
@akuckartz if we could re-write the JSON-LD spec, we'd do the exact same thing. The reason we used '@'-prefixed keywords was so that we could guarantee that the JSON-LD syntax wouldn't conflict with 99.999% of JSON markup out there. If we had standardized on 'id' or 'type' it would've started a firestorm of "well, my data has 'id' and 'type' all over it and therefore I can't use JSON-LD at all, you guys suck, I'm going to write a scathing blog post about how out of touch JSON-LD is now and share it with all my friends!" :P (a bit hyperbolic, but you get the idea).

akuckartz:
@msporny I get the idea, I already encountered significant amounts of Semaphobia ...

I will forward your suggestion to other communities. We will see how Hydra and schema.org react ;-)

vanthome:
@msporny I don't understand why the reason to go for`@xxx` you have outlined no longer exists?! I JSON-LD isn't out there for too long and I don't see that existing JSON based data has changed that much ever since.

lanthaler:
Hmm... if we would start from scratch again, then using id instead of @id might be an option.. but we can't. I thus think it is less confusing for users to simply stick to the standard and use those keywords. They will recognize them.. basically all blog posts, tutorials, presentations etc. that have been written use them. So should an upcoming W3C standard IMHO.

PR #381, opened 2015-01-25
Update example to point to dev context

jasnell:
Update example to point to the dev location for the context to avoid the warning. This ought to be temporary


msporny:
Thanks :)

jasnell:
Once the official URL is up I'll send along another pull request

PR #386, opened 2015-04-04
Center align "Playground" section on homepage

prtksxna:
The markup was using the `container` and `row` class on the same element. This patch moves the `row` class to the child `<div>` and moves the `offset` and `span` classes to the `<p>`.


msporny:
Thanks for the patch, @prtksxna ! :)

prtksxna:
> It's live on the site now.

Awesome!

PR #387, opened 2015-05-20
Remove normalization

gkellogg:
with reference to current location from the [latest location](http:json-ld.github.io/normalization/spec).

@manu, @dlongley sanity check please. Merge if satisified.


msporny:
Hey @gkellogg - I want to do a git filter-tree to save the history for the normalization stuff as it's ripped out and put in a separate repo. We try not to destroy history when we don't have to. Once I do that, I'll merge this commit.

gkellogg:
@msporny @dlongley I think we held off on this as you wanted to get history synced over to the new repository. However, actually committing this should affect that, if you ever want do do that in the future. This leaves one less thing for the CG to worry about. Without objection, I'll merge this PR.

gkellogg:
I believe we didn't complete this because of some desire to move the history over to https://github.com/json-ld/normalization. Note that even if it is committed, the history remains in the repo in case someone really cares about re-creating that.

I propose to close this and create a new PR that cleanly removes normalization from jsonld.org.

cc/ @msporny @davidlehn @dlongley @lanthaler

lanthaler:
+1

PR #392, opened 2015-08-28
update activity streams example

jasnell:
Use the official context document location and use Create instead of Post


lanthaler:
@jasnell could you please update http://www.w3.org/ns/activitystreams to return the correct media type, i.e. `application/ld+json` instead of `text/plain`. Thanks!

jasnell:
I'll submit a request, but it's not directly under my control. If it takes too long, we can point the example at the test endpoint again: `"@context": "http://as.jasnell.me"`

PR #403, opened 2015-11-24
Mention API Platform

dunglas:
Mention API Platform in PHP tools (a web framework based on Symfony to create JSON-LD enabled APIs).


davidlehn:
Does that project provide an alternate PHP implementation of JSON-LD or just use it?
If it's only using JSON-LD, it might be better to add it here:
https://github.com/json-ld/json-ld.org/wiki/Users-of-JSON-LD

dunglas:
It implements some part but mostly use it. Closing in favor of the wiki. Sorry for the noise.

PR #406, opened 2016-01-05
Test for parsing N-Quads containing comments

jindrichmynarz:
Test for <https://github.com/digitalbazaar/jsonld.js/pull/122>

### DIFF
--- a/test-suite/tests/fromRdf-0023-in.nq
+++ b/test-suite/tests/fromRdf-0023-in.nq
@@ -1,0 +1,6 @@
+<http://example.com/Subj1> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://example.com/Type> .
+# Comment
+<http://example.com/Subj1> <http://example.com/prop1> <http://example.com/Obj1> .
+<http://example.com/Subj1> <http://example.com/prop2> "Plain" . # Another comment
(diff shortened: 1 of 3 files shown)


gkellogg:
The fromRdf test suite isn't about parsing N-Quads, but about successfully generating JSON-LD given an N-Quads input. Testing comment parsing in N-Quads would be better served in https://github.com/w3c/rdf-tests test-suite, which already has such tests.

dlongley:
@jindrichmynarz, it's also possible to add [local tests](https://github.com/digitalbazaar/jsonld.js/tree/master/tests) to jsonld.js (not specific to json-ld.org's official test suite), it's just harder than it should be at the moment due to a very old testing framework.

jindrichmynarz:
@gkellogg, you're right that the test doesn't belong here. I see that <https://github.com/w3c/rdf-tests> test suite covers comments, but I don't know how it can be integrated with the jsonld.js test runner. @dlongley, is that possible?

In this case, I think it's simpler to just create a local test for jsonld.js.

dlongley:
@jindrichmynarz,

>  is that possible?

Yes, just again, a matter of time. We need to overhaul the test framework in jsonld.js and use something more modern -- and then also hook it up to things like rdf-tests where appropriate.

PR #411, opened 2016-02-28
fix some broken links in framing spec

tingletech:
so, a frame is like a template of what you want the final json-ld to look like?  Been banging my head on this spec with no luck.

do you happen to know if https://github.com/digitalbazaar/pyld and https://github.com/digitalbazaar/jsonld.js implement the same version of frames?

### DIFF
--- a/spec/latest/json-ld-framing/index.html
+++ b/spec/latest/json-ld-framing/index.html
@@ -113,7 +113,7 @@ enhances the functionality and interoperability of the Web.
 </p> -->
 </section>
 
-<section>
+<section id='introduction'>
 <h1>Introduction</h1>
 <p>A JSON-LD document is a representation of a directed graph. A single
   directed graph can have many different serializations, each expressing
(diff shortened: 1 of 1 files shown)


dlongley:
@tingletech,

> so, a frame is like a template of what you want the final json-ld to look like?

Yes, that's pretty close to the idea. It also acts like a filter. If you apply a frame to a set of data (a JSON-LD document), it will pick out the subjects/nodes that match the properties, types, ids, etc. in the template and fill it out according to its structure.

> Been banging my head on this spec with no luck.

Yeah, the spec is atrociously outdated and doesn't reflect current implementations.

> do you happen to know if https://github.com/digitalbazaar/pyld and https://github.com/digitalbazaar/jsonld.js implement the same version of frames?

Yes, they are the same. You can try out the jsonld.js version at the [JSON-LD playground](http://json-ld.org/playground).

lanthaler:
Thanks for fixing the links @tingletech

PR #417, opened 2016-04-12
Create LICENSE

dijonkitchen:


BigBlueHat:
👍

The web site states that it's "contents" are under the [Creative Commons CC0 Public Domain Dedication](http://creativecommons.org/about/cc0), but it's a bit unclear what the code is supposed to be licensed under. Is the code CC0 also?

Either way, it should be stated at the repository level ideally within a LICENSE file.

Also, it's [generally best to *not* use the CC0 license for code](https://opensource.org/faq#cc-zero), but to choose a more "seasoned" license such as the MIT, BSD, or Apache License 2.0.

All things you all likely all knew. 😸

msporny:
I agree with @BigBlueHat.

The original decision to CC-0 the work was made when the [CC-0-for-software finding](https://opensource.org/faq#cc-zero) hadn't been made yet.

The site content is CC-0, we can make the code BSD or MIT, don't have a strong preference on which one. We will also want to state that software libraries that are included with different licenses keep the licenses that were assigned to them originally (clearly, we can't switch the jsonld.js library license just because we use it here). Can you update this PR to say that, @dijonkitchen? We'll merge after that change.

gkellogg:
Another alternative would be to split licensing. For example, there is a particular license for W3C test suites: https://www.w3.org/Consortium/Legal/2008/04-testsuite-copyright.html. And, the specs themselves have their own license. We should probably describe that the BSD/MIT license does not apply to the specifications and test suite.

gkellogg:
W3C Document License: https://www.w3.org/Consortium/Legal/2015/doc-license

gkellogg:
We should also have a "contributing.md" file, which will show on any pull request. CSVW used the following: https://github.com/w3c/csvw/blob/gh-pages/contributing.md.

BigBlueHat:
A `NOTICES` file generally serves the purpose @msporny describes. Additionally, I'd be happy to help with refactoring some of the site code to a) be more static (via jekyll if that works for you all) and  b) more modular (via browserify perhaps).

The goal being that the pieces not under the "repo-level" license could clearly contain (or reference) their own licensing information. AFAIK the license declaration "nearest" to the actual content is the one that prevails. Additionally, having the license declaration being made as near to (and as obviously on behalf of) the original author(s) the better (hence the package-based remix suggestion).

Also 👍 to the `contributing.md` file suggestion and using the CSVW one as a foundation.

dijonkitchen:
New PR in #456 based on the comments above since this went stale and I deleted my old fork and branch.

PR #421, opened 2016-06-20
Fix json-schema validation

andersoncardoso:
Hi, I believe I found two small issues on the jsonld-schema.

###  issue 1

A hidden U+FEFF char on the begining of the file was breaking some parsers, for example in ruby: 

```
content  = File.read('./schemas/jsonld-schema.json')
JSON.parse content  # raise an Error
```

### issue 2

The schema definition seems to have a validation mistake. I.e, almost any invalid json-ld would pass.

This happens because we are using a `anyOf` clause on the definitions, so if we provide an invalid `@context`, the validator will check for the next and one of them almost always pass. This way the json would be invalid only if context, common and graph are invalid at the same time. 

To fix, we just have to change to a `allOf` clause, i.e, each of the defintions must be valid at the same time.

To ilustrate this issue, I'm providing a small test below (it uses the `json-schema` rubygem):

```ruby
require 'json-schema'

puts JSON::Validator.validate('../json-ld.org/schemas/jsonld-schema.json', {}) ? 'ok' : 'NOPE!'
puts !JSON::Validator.validate('../json-ld.org/schemas/jsonld-schema.json', {'@id': true, '@context': 42}) ? 'ok' : 'NOPE!'

```

I hope this can help someone,

tks

### DIFF
--- a/schemas/jsonld-schema.json
+++ b/schemas/jsonld-schema.json
@@ -79,7 +79,7 @@
         }
     },
 
-    "anyOf": [
+    "allOf": [
         { "$ref": "#/definitions/context" },
         { "$ref": "#/definitions/graph" },
         { "$ref": "#/definitions/common" }
--- a/schemas/jsonld-schema.json
+++ b/schemas/jsonld-schema.json
@@ -44,8 +44,7 @@
                 },
                 "@type": {
                     "description": "Used to set the data type of a node or typed value.",
-                    "type": ["string", "null"],
-                    "enum": ["@id"]
+                    "type": ["string", "null"]
                 },
                 "@container": {
                     "description": "Used to set the default container type for a term.",
--- a/schemas/jsonld-schema.json
+++ b/schemas/jsonld-schema.json
@@ -1,7 +1,7 @@
-﻿{
+{
     "title": "Schema for JSON-LD",
     "$schema": "http://json-schema.org/draft-04/schema#",
-    
+
     "definitions":{
         "context": {
             "additionalProperties": true,


gkellogg:
The odd character in the file seems persistent, and may need the file to be replaced; we can see post-merge.

The change to `allOf` from `anyOf` seems problematic, as this would seem to require an `@graph` property, which does not necessarily need to be in a JSON-LD file. Otherwise, 👍 .

The schema was originally created by @madskristensen, who may want to weigh in.

andersoncardoso:
Hi @gkellogg, thanks for your fast reply =)

Maybe I'm missing something, but I believe the `@graph` property is not required, since any of the properties are on a `required` list.
Almost all json data I'm validating now with this schema doesn't have the `@graph` property and I haven't got any error so far.

for the hidden char, I've tried to re-clone my fork from scratch an then:

```bash
git checkout master
ruby -e "p File.read('./schemas/jsonld-schema.json').match(/\ufeff/) ? 'BOM' : 'ok'"
# >> "BOM"

git checkout fix-jsonld-schema
ruby -e "p File.read('./schemas/jsonld-schema.json').match(/\ufeff/) ? 'BOM' : 'ok'"
# >> "ok"
```

so it doesn't seems to be persistent, you are still getting this char on the file?

If I can help in any way to explore this further, let me know.

gkellogg:
> Maybe I'm missing something, but I believe the @graph property is not required, since any of the properties are on a required list.

Sorry, you're right, the fact that it's not on the `required` list should be sufficient. I'm only somewhat familiar with JSON Schema, so I'm sure I have some miss-understandings.

As a sanity test, it might be good to pass this against all of the (valid) JSON-LD files in the [/test-suite/tests](https://github.com/json-ld/json-ld.org/tree/master/test-suite/tests) directory to ensure these all validate with the schema, and if not, if that's intended.

PR #423, opened 2016-08-03
Playground button css fixed

ivikash:
Hey, 

The text in button's of playground didnt feel right, with this PR I intend to fix the padding issue

> Before
> screen shot 2016-08-03 at 3 21 00 pm
> 
> After
> screen shot 2016-08-03 at 3 28 31 pm

### DIFF
--- a/playground/playground.css
+++ b/playground/playground.css
@@ -7,6 +7,9 @@
   background-color: #fff;
 }
 
+.btn-pill{
+  padding: 0.6em 0.7em;
+}
 
 .btn.resizer{
   cursor: row-resize;
--- a/playground/playground.css
+++ b/playground/playground.css
@@ -7,7 +7,6 @@
   background-color: #fff;
 }
 
-
 .btn.resizer{
   cursor: row-resize;
   padding: .5px;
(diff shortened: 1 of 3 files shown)


dlongley:
If this kind of problem needs to be fixed, it should have some kind of bootstrap fix; I don't think we should start doing custom CSS for it.

ivikash:
Infact at first I thought the same. Lets make it generic. but then the form inside gist started breaking. So what I will do is, I will fix those in bootstrap.css itself :)

dlongley:
If we need to upgrade bootstrap, that's fine, but let's not make changes to bootstrap.css unless they are the types of fixes that would be submitted to that project directly. We should be able to use vanilla bootstrap properly to get a nice looking site.

ivikash:
Updated changes
screen shot 2016-08-03 at 11 48 06 pm

playground/playground.css
```diff
@@ -60,14 +56,15 @@
...
-}
-
 #page-playground .read-only {
   opacity: .6;
 }
 
+#page-playground .btn {
+  height: 30px;
+  padding: 0 0.7em;
+}
```
dlongley: Ideally, there wouldn't be anything custom required here at all. If the buttons still look wrong, we must not be doing something quite right with bootstrap. There default theme looks just fine (and we're trying to use that, AFAIK).

gkellogg:
This PR is languishing, we should either commit to close.

cc/ @msporny @davidlehn @dlongley @lanthaler.

Issue #110, opened 2012-04-26
More specific frame matching


gkellogg:
Currently, framing allows you to select subject definitions based on @type matching or duck typing for included properties. It allows value properties to be explicitly matched based on defining the property and excluding things that are undefined, but it does not allow you to be more specific about the types of values selected.

For example, DBPedia creates output in many languages, and it would be reasonable to frame to get just a specific language. This could be supported by adding more specific value matching. Consider the following:

Document:

```
{
  "rdfs:comment": [
    {"@value": "comment in english", "@language": "en"},
    {"@value": "commentar auf deutsch", "@language": "de"},
    {"@value": "日本語でのコメント", "@language": "ja"},
  ]
}
```

Frame:

```
{
  "@context": {},
  "rdfs:comment": {"@value": {}, "@language": "en"}
}
```

This might select only those values that match the specific language. Or, using a {} wildcard, only those values that have a defined language. The same could be used for @type: {}, when used in conjunction with@value. (We could also consider @language: false or @type: false to indicate that there is no language or type.

Using the array syntax, multiple values could be selected:

```
{
  "@context": {},
  "rdfs:comment": [
    {"@value": {}, "@language": "en"}
    {"@value": {}, "@language": "de"}
  ]
}
```

This could be extended to say that for a property, the {} wildcard only selects resources, and to select a value requires an explicit {"@value": {}} to be a value. Personally, I think that if you use {"@type": {}}, it's clear that you only want values which are resources, and that {} can remain a wildcard matching both values and resources.

gkellogg:
See "Frame Matching Algorithm" in https://rawgit.com/json-ld/json-ld.org/issue-110-frame-matching/spec/latest/json-ld-framing/index.html

Issue #424, opened 2016-09-08
JSON framing : support for multiple @id


abrisse:
JSON-LD framing should support multiple `@id` in order to "hook" multiple IRIs that have nothing in common in kind of relations (`@type` or whatsoever).

A solution would be to be able to accept arrays for `@id`.

``` json{
  "@context": {
    "dc": "http://purl.org/dc/elements/1.1/",
    "ex": "http://example.org/vocab#"
  },
  "@id": [
    "http://example.org/library",
    "http://example.org/kitchen"
  ]
}
```

Currently an error is returned : 

> Invalid JSON-LD syntax; `@id` value must be a string or an object.

Note: `@type` should accept multiple values too

Issue #377, opened 2014-11-26
Update framing spec to use new embed API


dlongley:
A new experimental embed API has been implemented in jsonld.js. Instead of using `"@embed": true/false`, three options for the `@embed` directive are available: `@always`, `@never`, and `@last`. The current framing behavior for `"@embed": true` matches that of `@last` which is the default. There's backwards compatibility support in jsonld.js for the previous `@embed` boolean flags.

I think this is a pretty good way forward for more powerful embedding w/framing. I recently simplified parts of the framing algorithm in jsonld.js related to embedding as well -- so I decided to make an issue here for tracking the need to update the framing spec. I think when we do an update, we should include this new API and we can look at the jsonld.js implementation to help simplify the algorithm prose. Thoughts?

See: https://github.com/digitalbazaar/jsonld.js/pull/68

Addresses: #119.

kevinSuttle:
@dlongley This is fascinating. I would love to hear your thoughts on how something like this would work on the effort currently underway: https://github.com/schemaorg/schemaorg/issues/created_by/kevinSuttle

kevinSuttle:
Also, is this issue still open? Looks like https://github.com/digitalbazaar/jsonld.js/pull/68 is closed and https://github.com/json-ld/json-ld.org/issues/119 probably should be.

lanthaler:
Yes, this issue is still open. We aren’t working on the specs at the moment.

gkellogg:
This is addressed in 5a3e50610ef7c1fc08f92a2a5b393ab9741c0c11. However, I think there's more room for discussion on @embed, and order dependency in general.

Issue #362, opened 2014-09-10
Spec Enhancement: Complex JSON-LD Framing support


pkuyken:
Currently JSON-LD Framing supports the "@explicit" : true option to allow only properties explicitly stated in the frame to be included in the framed graph output. This approach does not allow for a stricter framing that will only allow properties _and_ "rdf:type"s that are explicitly defined to be used in the framed output graph.  

For example given the following JSON-LD input:

``` json
[
  {
    "@id": "http://example.com/resource/initial",
    "@type": [
      "http://example.com/sample.owl#Resource",
      "http://example.com/hypermedia.owl#InitialState"
    ],
    "http://example.com/sample.owl#startDate": [
      {
        "@value": "2002-05-30T09:00:00",
        "@type": "http://www.w3.org/2001/XMLSchema#date"
      }
    ]
  }
]
```

And the following JSON-LD Frame:

``` json
{
  "@context": {
    "ex": "http://example.com/sample.owl#",
    "foaf": "http://xmlns.com/foaf/0.1/"
  },
  "@type": "ex:Resource",
  "@explicit": "@value",
  "foaf:name": {}
}
```

The desired ouput should be:

``` json
{
  "@context": {
    "ex": "http://example.com/sample.owl#",
    "foaf": "http://xmlns.com/foaf/0.1/"
  },
  "@graph": [
    {
      "@id": "http://example.com/resource/initial",
      "@type": "ex:Resource",
      "foaf:name": null
    }
  ]
}
```

I propose enhancing the "@explicit" option to support a wider variety of framing options.  In particular an option "@value" where the frame would only include those properties _and_ values that are explicitly mentioned in the frame.

This is similar to the approach used with @container and the proposed enhancement for @embed.

gkellogg:
Note that proposed CG updates to the framing spec updates `@embed` support: 5a3e50610ef7c1fc08f92a2a5b393ab9741c0c11. However, this use case would seem to address the desire to _only_ include values that are specified in the frame, in this case, limit the `@type` emitted to the intersection of the types in the frame and the types in the source. I don't see `@value` myself, but perhaps something like `@explicit: @frame` (take values from the frame, not the source, or `@explicit: @intersect` to take the intersection of values from the frame and the source.

Note that this is rather particular to `@type`, as other properties take just a single value, used for embedding or simple detection of property presence, rather than property value matching.

gkellogg:
See also #110.

gkellogg:
Closed by PR #437.

PR #437, opened 2016-11-08
Issue 110 frame matching

gkellogg:
This completes the work on issue #110. The exception is for question 3:

> Do we interpret `[ a, b ]` the value of subject's property should be `a` OR `b`?

This is different from any other framing (except for multiple values of `@type` and `@language` within a value pattern), and require more iteration during matching and framing.

This PR also updates the API document with framing-specific algorithm updates.

Until merged, see [json-ld-framing](https://rawgit.com/json-ld/json-ld.org/issue-110-frame-matching/spec/latest/json-ld-framing/index.html) and [json-ld-api](https://rawgit.com/json-ld/json-ld.org/issue-110-frame-matching/spec/latest/json-ld-api/index.html)


gkellogg:
When accepted, this will close #110, #362, #377, and #424.

spec/latest/common/terms.html
```diff
@@ -146,11 +146,9 @@
       tag as defined by [[!BCP47]]. The language tag MUST be well-formed according to
       <a href="http://tools.ietf.org/html/bcp47#section-2.2.9">section 2.2.9 Classes of Conformance</a>
       of [[!BCP47]], and is normalized to lowercase.</dd>
-    <dt><dfn data-lt="lists">list</dfn></dt><dd>
-      A <a>list</a> is an ordered sequence of <a>IRIs</a>,
-      <a>blank nodes</a>, and <a>JSON-LD values</a>.</dd>
     <dt><dfn>Linked Data</dfn></dt><dd>
       A set of documents, each containing a representation of a <a>linked data graph</a>.</dd>
+      <a>blank nodes</a>, and <a>JSON-LD values</a>.</dd>
```
dlongley: This looks like it got transposed ... and should be down lower in the `list` definition.

spec/latest/json-ld-api/index.html
```diff
@@ -1187,7 +1187,21 @@
         To begin, the <a>active property</a> is set to <code>null</code>,
         and <em>element</em> is set to the <a>JSON-LD input</a>.</p>
 
-      <ol class="algorithm">
+      <p>The algorithm also performs processing steps specific to expanding
+        a <a>JSON-LD Frame</a>. For a <a>frame</a>, the <code>@id</code> and
+        <code>@type</code> properties can accept an array of <a>IRIs</a> or
+        an empty <a>dictionary</a>. The properties of a <a>value object</a> can also
+        accept an <a>array</a> of <a>strings</a>, or an empty <a>dictionary</a>.
```
dlongley: Do we also need to accept JSON numbers, booleans, etc. for value objects?

spec/latest/json-ld-framing/index.html
```diff
@@ -484,8 +474,9 @@
                     as the sole member of a new <em>subjects</em> <a>array</a>,
                     the first value from <em>property</em> in <em>frame</em> as <em>frame</em>,
                     <em>output</em> as <em>parent</em>, and <em>property</em> as <em>active property</em>.</li>
-                  <li>Otherwise, append a copy of <em>item</em> to <a>active property</a> in
-                    <em>output</em>.</li>
+                  <li>Otherwise, if item matches using the <a href="#value-matching">Value matching algorithm</a>,
+                    where the <em>pattern</em> is taken from the first value from <em>property</em> in <em>frame</em>,
+                    append a copy of <em>item</em> to <a>active property</a> in <em>output</em>.</li>
```
dlongley: I think we may need to discuss what's going on here with using the first value as a pattern (to match, I presume in the value matching algorithm), but that can happen post merging this PR. This is at least a start for handling value matching, which we had nothing for previously.

spec/latest/json-ld-framing/index.html
```diff
@@ -552,34 +543,109 @@
...
+  <p>The Frame Matching Algorithm is used as part of the <a href="#framing-algorithm">Framing algorithm</a>
+    to determine if a particular <a>node object</a> matches the critera set in a <a>frame</a>.
+    In general, a node object matches a frame if it meets the matches on <code>@type</code>,
+    or <code>@id</code>,
+    or if it matches given one of several different properties (or all properties, if the
+    <a>require all flag</a> is present.).</p>
+
+  <p class="note">As matching is performed on expanded node objects, all values will be in the form of an array.</p>
+
+  <p>Node matching uses a combination of JSON constructs to match <em>any</em>, <em>no</em>, or <em>some</em> specific values:</p>
```
dlongley: Use of `no` here is a bit strange but I don't know if something like `zero` is better. Just noting that it was an odd read.

spec/latest/json-ld-framing/index.html
```diff
@@ -552,34 +543,109 @@
...
+  <dl>
+    <dt><code>[]</code> (<code><dfn>match none</dfn></code>)</dt>
+    <dd>An empty array matches no values, or a value which is, itself, an empty array.</dd>
```
dlongley: +1

spec/latest/json-ld-framing/index.html
```diff
@@ -552,34 +543,109 @@
...
+    <dt><code>{}</code> (<code><dfn>wildcard</dfn></code>)</dt>
+    <dd>An array containing an empty object
+      (after excluding the framing properties <code>@default</code>, <code>@embed</code>, <code>@explicit</code>, <code>@omitDefault</code>, and <code>@requireAll</code>)
+      matches any value that is present, and does not match if there are no values.</dd>
```
dlongley: +1. I don't know if this is true or not yet, but it would be nice if we could simplify by saying "after excluding any framing keywords", rather than having to enumerate them.

spec/latest/json-ld-framing/index.html
```diff
@@ -552,34 +543,109 @@
...
+    <dt><code>[<a>IRI</a>+]</code></dt>
+    <dd>One or more strings in the form of an IRI, used for matching on <code>@type</code> and <code>@id</code>.</dd>
```
dlongley: We might want to mention that the matching uses "or" semantics here.

spec/latest/json-ld-framing/index.html
```diff
@@ -552,34 +543,109 @@
...
+    <dt><code>[<a>node object</a>]</code> (<code><dfn>node pattern</dfn></code>)</dt>
+    <dd>A non-empty <a>node object</a>, used to match specific values using recursive node matching.</dd>
```
dlongley: Should this be called a node object or a frame/subframe?

spec/latest/json-ld-framing/index.html
```diff
@@ -552,34 +543,109 @@
...
+    <dt><code>[<a>value object</a>]</code> (<code><dfn>value pattern</dfn></code>)</dt>
+    <dd>A <a>value object</a>, used to match a specific value. Within a <a>value object</a>,
+      the values for <code>@value</code>, <code>@type</code>, and <code>@language</code>
+      may also be an array of one or more <a>string</a> values.</dd>
```
dlongley: Is this actually limited to a single value object -- given the above reference to using only the first value as the pattern?

spec/latest/json-ld-framing/index.html
```diff
@@ -552,34 +543,109 @@
...
+  </dl>
+
+<p>The frame matching algorithm takes the <a>framing state</a> (<em>state</em>),
+  a list of subjects to match from the <a>map of flattened subjects</a> (<em>subjects</em>),
+  a <a>frame</a> to match against (<em>frame</em>), and the <em>requireAll</em> flag
+  and returns a list of matched subjects by filtering each <em>node</em> in <em>subjects</em> as follows:</p>
+
+<p>Frame matching follows a cascade, first attempting to match on a particular <code>@id</code>, then
+  a particular <code>@type</code> (or lack of <code>@type</code>), then by matching on any or all
+  of a set of properties, if neither <code>@id</code>, nor <code>@type</code> are in the frame.</p>
```
dlongley: We might want to refer to this as an order of precedence. That being said, I agree with what's here.

spec/latest/json-ld-framing/index.html
```diff
@@ -552,34 +543,109 @@
...
+
+<ol class="algorithm">
+  <li><em>Node</em> matches if it has an <code>@id</code> property including any <a>IRI</a> in the <code>@id</code> property in <em>frame</em>.</li>
```
dlongley: Note: it may not be clear to people that: `"@id": []` won't work for capturing blank nodes ... because the flattened subject map will assign a blank node to everything (IIRC).

spec/latest/json-ld-framing/index.html
```diff
@@ -552,34 +543,109 @@
...
+  <li>If <em>requireAll</em> is <strong>true</strong>, <em>node</em> matches if all non-<a>keyword</a> properties (<em>property</em>) in <em>frame</em> match any of the following conditions.
+    Or, if <em>requireAll</em> is <strong>false</strong>, if any of the non-<a>keyword</a> properties (<em>property</em>) in <em>frame</em> match any of the following conditions.
+    For the <em>values</em> of each <em>property</em> from <em>frame</em> in <em>node</em>:
+    <ol class="algorithm">
+      <li>If <em>property</em> is <code>@type</code>:
+        <ol class="algorithm">
+          <li><em>Property</em> matches if the <code>@type</code> property in frame includes any <a>IRI</a> in <em>values</em>.</li>
+          <li>Otherwise, <em>property</em> matches if <em>values</em> is not empty and the <code>@type</code> property in <em>frame</em> is <code><a>wildcard</a></code>.</li>
+          <li>Otherwise, <em>property</em> matches if <em>values</em> is empty and the <code>@type</code> property in <em>frame</em> is <code><a>match none</a></code>.</li>
+          <li>Otherwise, <em>property</em> does not match.</li>
```
dlongley: LGTM

spec/latest/json-ld-framing/index.html
```diff
@@ -552,34 +543,109 @@
...
+        </ol>
+      </li>
+      <li>Otherwise, the value of <em>property</em> in <em>frame</em> MUST be empty, or an array
+        containing a valid <a>frame</a>.</li>
```
dlongley: Or what? Throw an error?

spec/latest/json-ld-framing/index.html
```diff
@@ -552,34 +543,109 @@
...
+      <li><em>Property</em> matches if <em>values</em> is empty, or non existant,
+        the value of <em>property</em> in <em>frame</em>
+        is a <a>dictionary</a> containing only the key <code>@default</code> with any value,
+        and any other property in <em>node</em> has a non-default match.</li>
+      <li><em>Node</em> does not match if <em>values</em> is not empty and the value of <em>property</em> in <em>frame</em> is <code><a>match none</a></code>, and further matching is aborted.</li>
```
dlongley: Switching between "property matches" and "node matches" was a bit confusing to me here. May be easier to understand when viewing w/styling.

spec/latest/json-ld-framing/index.html
```diff
@@ -552,34 +543,109 @@
...
+  </li>
+</ol>
+</section>
+
+<section id="value-matching">
+  <h2>Value Pattern Matching Algorithm</h2>
+
+  <p>The Value Pattern Matching Algorithm is used as part of the <a href="#framing-algorithm">Framing</a>
+    and <a href="#frame-matching">Frame Matching</a> algorithms. A value object
+    matches a value pattern using the <code><a>match none</a></code>, <code><a>wildcard</a></code>
```
dlongley: There are only two patterns listed here, but a comma is used. Is there supposed to be a third?

dlongley:
I've finished my review. I had a few minor comments and some editorial stuff to address, but overall, I think this is good and we should merge it. I need to get some implementation experience with the algorithm and see if the test inputs and outputs seem like they make sense.

There are some additions to the existing algorithm (e.g. value matching) but I only saw one breaking change -- which was how strict typing vs. duck typing are treated. I tentatively think this breaking change is for the better, so we should merge this. I'd like to confirm my feeling on that with some implementation experience when I can.

This is awesome. Thanks, Gregg!

spec/latest/json-ld-api/index.html
gkellogg: Yes, to be complete, we need to accept an array composed of anything that can be a value of `@value` in the JSON-LD grammar.

spec/latest/common/terms.html
gkellogg: What was transposed? I wanted to keep the terms in alphabetical order.

spec/latest/json-ld-api/index.html
```diff
@@ -3306,8 +3345,6 @@
                 <li>While <em>property</em> equals <code>rdf:rest</code>,
                   the <a>array</a> value of the member of <em>node usages map</em> associated with the <code>@id</code>
                   member of <code>node</code> has only one member,
-                  the value associated to the <code>usages</code> member of <em>node</em> has
-                  exactly 1 entry,
```
gkellogg: While I would like to revisit these algorithms with an eye towards removing implementation details and focusing on the intent of the algorithms, this seems to have been an accidental change, which I'll reverse.

spec/latest/json-ld-framing/index.html
```diff
@@ -301,7 +305,10 @@
       <a href="../json-ld-api/#compaction">Compacting</a>.</dd>
     <dt><code>@omitDefault</code></dt>
     <dd>Used in <a href="#framing">Framing</a> to override the
-      value of <a>omit default flag</a> within a specific frame.</dd>
+      value of <a>omit default flag</a> within a specific frame.
+      <div class="issue">Is this flag really useful? Easier to simply not have a default value,
+        if it shouldn't match.</div>
```
gkellogg: It's confusing naming. If a property value has `@omitDefault: true`, it will not generate a property with `null` in the output, or use the value of `@default`. Otherwise, we could just not put the `@default` in in the first place. An alternative would be to remove `@omitDefault` and _not_ add missing property values to the output without `@default`. We could then allow `@default: @null` to explicitly give a `null` value in the output.

spec/latest/json-ld-framing/index.html
gkellogg: We had discussed having multiple patterns in a frame, but I felt this to be a big step; we can indeed discuss later on.

spec/latest/json-ld-framing/index.html
gkellogg: `zero` is probably better.

spec/latest/json-ld-framing/index.html
gkellogg: Yes, I think we can simplify it.

spec/latest/json-ld-framing/index.html
gkellogg: Adding ", which allows a match on <em>any</em> of the listed IRIs."

spec/latest/json-ld-framing/index.html
gkellogg: We might want to define either the term subframe, or frame object so that we can use it consistently; this does tend to come up many different times. _Frame object_ could be defined as a kind of _node object_ extended with framing semantics.

spec/latest/json-ld-framing/index.html
gkellogg: There is only one _value pattern_, it's `@value` property may have multiple values for matching. Allowing multiple _value patterns` would make frame validation more difficult, and processing would need to do more iteration, so I've avoided this for now.

spec/latest/json-ld-framing/index.html
gkellogg: 👍

spec/latest/json-ld-framing/index.html
gkellogg: Should be any IRI or Blank Node in the `@id` property, with a note that flattening has the effect of ensuring that all _node objects_ have an `@id`, thus `@id: []` will match zero node objects.

spec/latest/json-ld-framing/index.html
```diff
@@ -552,34 +543,109 @@
...
+
+<p>Frame matching follows a cascade, first attempting to match on a particular <code>@id</code>, then
+  a particular <code>@type</code> (or lack of <code>@type</code>), then by matching on any or all
+  of a set of properties, if neither <code>@id</code>, nor <code>@type</code> are in the frame.</p>
+
+<ol class="algorithm">
+  <li><em>Node</em> matches if it has an <code>@id</code> property including any <a>IRI</a> in the <code>@id</code> property in <em>frame</em>.</li>
+  <li><em>Node</em> matches if frame has no non-<code>keyword</code> properties.</li>
+  <li>If <em>requireAll</em> is <strong>true</strong>, <em>node</em> matches if all non-<a>keyword</a> properties (<em>property</em>) in <em>frame</em> match any of the following conditions.
+    Or, if <em>requireAll</em> is <strong>false</strong>, if any of the non-<a>keyword</a> properties (<em>property</em>) in <em>frame</em> match any of the following conditions.
```
gkellogg: This is the conclusion I came to, but my understanding of the interplay between `@type` and `requireAll` was a bit different. We can re-visit after implementation experience.

spec/latest/json-ld-framing/index.html
gkellogg: The change is because we match a specific property rather than the whole node. If `requireAll` is set, all properties must have matched. Otherwise, any property. A mis-matching property does not necessarily cause the node match to fail.

spec/latest/json-ld-framing/index.html
```diff
@@ -552,34 +543,109 @@
...
+    <li>Let <em>v2</em>, <em>t2</em>, and <em>l2</em> be the values of <code>@value</code>, <code>@type</code>, and <code>@language</code> in <em>value pattern</em>, or <strong>null</strong> if none exists.</li>
+    <li><em>Value</em> matches <em>pattern</em> when pattern is <code><a>wildcard</a></code>, or:
+      <ol class="algorithm">
+        <li><em>v1</em> is in <em>v2</em>, or <em>v1</em> is not <strong>null</strong> and <em>v2</em> is <code><a>wildcard</a></code>, and</li>
+        <li><em>t1</em> is in <em>t2</em>, or <em>t1</em> is not <strong>null</strong> and <em>t2</em> is <code><a>wildcard</a></code>, or <strong>null</strong>, or <em>t1</em> is <code>null</code> and <em>t2</em> is <strong>null</strong> or <code><a>match none</a></code>, and</li>
+        <li><em>l1</em> is in <em>l2</em>, or <em>l1</em> is not <strong>null</strong> and <em>l2</em> is <code><a>wildcard</a></code>, or <strong>null</strong>, or <em>l1</em> is <code>null</code> and <em>l2</em> is <strong>null</strong> or <code><a>match none</a></code>.</li>
+      </ol>
```
gkellogg: 👍

spec/latest/common/terms.html
dlongley: Line 151 here is the second part of the list definition. It just didn't get properly copied down below to where you moved the list stuff.

spec/latest/common/terms.html
```diff
@@ -159,6 +157,8 @@
       A <a>linked data graph</a> is a generalized representation of a
       <cite><a href="http://www.w3.org/TR/rdf11-concepts/#dfn-rdf-graph">RDF graph</a></cite>
       as defined in [[!RDF-CONCEPTS]].</dd>
+    <dt><dfn data-lt="lists">list</dfn></dt><dd>
+      A <a>list</a> is an ordered sequence of <a>IRIs</a>,
```
dlongley: That list description should be pasted down here following line 161.

spec/latest/common/terms.html
dlongley: I added a note below around line 161 where it should go.

spec/latest/json-ld-framing/index.html
dlongley: I'm not opposed to changing this behavior around, especially if it is confusing. We should probably add a few examples to the test suite as use cases for this... just want to make sure it's easy for people to configure when defaults appear and when they do not with a reasonable level of granularity. This shouldn't hold up the PR, it's just a discussion/implementation experience thing to consider later.

spec/latest/json-ld-framing/index.html
dlongley: Ok. Fine for now, but we may want to be more explicit with telling people to throw errors (and which "type").

spec/latest/common/terms.html
gkellogg: Got it, thanks.

PR #438, opened 2016-11-20
Issue 118 graph framing

### LINKED ISSUE #118: How is @graph going to be used/supported in framing?
We haven't discussed yet how `@graph` can be used in framing. I think we should do the following:
- if `@graph` isn't used in the frame (till that point), the input (subject map) to the framing algorithm is the merge of all graphs in the input document
- the first use of `@graph` without an `@id` will use the default graph (this has to be specified a bit better)
- all other uses of `@graph` will use the specified named graph (what happens if no `@id` was set?)

Thoughts?

gkellogg:
This PR closes #118, adding support for named graphs when framing, and allows framing to use either the merged graph or the default graph. It also adds an algorithm for creating the merged node set.

### DIFF
--- a/spec/latest/json-ld-api/index.html
+++ b/spec/latest/json-ld-api/index.html
@@ -2923,6 +2923,34 @@
       </ol>
     </section>
   </section> <!-- end of Generate Blank Node Identifier -->
+
+  <section>
+    <h3>Merge Node Maps</h3>
+    <p>This algorithm creates a new map of <a>subjects</a> to <a>nodes</a> using all graphs
+      contained in the <em>graph map</em> created using the <a href="#node-map-generation">Node Map Generation algorithm</a>
+      to create merged <a>node objects</a> containing information defined for a given <a>subject</a>
+      in each graph contained in the <em>node map</em>.</p>
+
+    <ol class="algorithm">
+      <li>Create <em>result</em> as an empty <a>dictionary</a></li>
+      <li>For each <em>graph name</em> and <em>node map</em> in <em>graph map</em>
+        and for each <em>id</em> and <em>node</em> in <em>node map</em>:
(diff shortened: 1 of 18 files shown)


gkellogg:
See the [formatted document](https://rawgit.com/json-ld/json-ld.org/issue-118-graph-framing/spec/latest/json-ld-framing/index.html) for easier review.

dlongley:
I don't have much time to do a thorough review, but it's based off of the experimental work I was doing -- and making sure this supports the main use cases will require implementing it anyway. I support merging this PR.

The main case I'm concerned with is being able to produce a structure like this:

```js
{
  "@context": "...",
  "id": "person1",
  "type": "IdentityProfile",
  "credential": [{
    "@graph": {
      "id": "cred1",
      "claim": {
        "id": "person1",
        "ageOver": 21
      },
      "signature": { ... }
    }, {
    "@graph": {    
      "id": "cred2",
      "claim": {
        "id": "person1",
        "passport": { ... },
      },
      "signature": { ... }
    }
  }]
}
```

This use case is about being able to receive an "IdentityProfile" in any format and then restructure it via framing for signature verification. The "IdentityProfile" consists of a set of credentials, each one containing verifiable claims about the subject of the IdentityProfile -- and the integrity of each must be preserved as an independent graph to enable proper verification of the signatures. Once the signatures have been checked, regular (merged) framing could also be used on this input to produce a utilitarian, merged view of the "person1" node containing all of the various properties from the verifiable claims.

gkellogg:
If you have some concrete input data and frame, we can add this as a test to make sure it works. Post a gist with some input and I can create a framing test for it.

dlongley:
@gkellogg -- I'll post a gist of some concrete input when I get a chance.

PR #442, opened 2016-12-06
Add JSON-LD verification via BTC public key

harlantwood:
We have signature validation with bitcoin keys [working and live on a demo server](http://thawing-inlet-70039.herokuapp.com/playground/).  

**We should wait for the following to be merged before merging this PR:**

- [x] https://github.com/digitalbazaar/jsonld-signatures/pull/10
- [x] https://github.com/digitalbazaar/jsonld-signatures/pull/16
- [x] https://github.com/digitalbazaar/jsonld-signatures/pull/17

Once https://web-payments.org/contexts/security-v1.jsonld contains EcdsaKoblitzSignature2016 we can: 

- [x] Change https://comakery.github.io/json-ld.org/schemas/security-v1-patch.jsonld to https://w3id.org/security/v1 within this PR

Nice to have but probably not critical:

https://github.com/web-payments/web-payments.org/issues/41

cc @ChristopherA @msporny

### DIFF
--- a/playground/index.html
+++ b/playground/index.html
@@ -252,13 +252,13 @@ yT2IMAWbY76Bmi8TeQJAfdLJGwiDNIhTVYHxvDz79ANzgRAd1kPKPddJZ/w7Gfhm
             </div>
 
             <div id="privatekey-koblitz-div" class="span6">
-              <h3>Bitcoin Private Key (ECDSA Koblitz)</h3>
+              <h3>Bitcoin (ECDSA Koblitz) Private Key for Signing</h3>
               <textarea id="privatekey-koblitz" class="compressed process span6 codemirror-input"
-                placeholder="Enter your Bitcoin (koblitz) private key here..." rows="3">L4mEi7eEdTNNFQEWaa7JhUKAbtHdVvByGAqvpJKC53mfiqunjBjw</textarea>
+                placeholder="Enter your Bitcoin (Koblitz) private key here..." rows="3">L4mEi7eEdTNNFQEWaa7JhUKAbtHdVvByGAqvpJKC53mfiqunjBjw</textarea>
 
-              <h3>Bitcoin Public Key (ECDSA Koblitz)</h3>
+              <h3>Bitcoin (ECDSA Koblitz) Public Key for Verification</h3>
               <textarea id="publickey-koblitz" class="compressed process span6 codemirror-input"
-                placeholder="Enter your Bitcoin (koblitz) public key here..." rows="3">1LGpGhGK8whX23ZNdxrgtjKrek9rP4xWER</textarea>
+                placeholder="Enter your Bitcoin (Koblitz) public key here..." rows="3">1LGpGhGK8whX23ZNdxrgtjKrek9rP4xWER</textarea>
 
               <div class="koblitz-verification"></div>
             </div>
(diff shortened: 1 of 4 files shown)


harlantwood:
@ChristopherA @msporny @aquabu @gkellogg this is ready to merge IMO.  Please take a look at the demo: http://thawing-inlet-70039.herokuapp.com/playground/ and/or the code, and let's give it some 👍  or 👎 .  Thanks!

msporny:
Works for me, fine with it getting merged. 

I have the following requests for future discussion:

1. We need to figure out how to make the UI for all of this more intuitive. It'll help if/when we launch a digital verification playground. 
2. There is no way to put in some signed JSON-LD and see if the signature checks out. The Koblitz stuff only checks to see if the displayed information is valid (which is fine, but the DV Playground should probably just have a signature checking mechanism).
3. The key format is strange because it 1) doesn't allow lookups, 2) doesn't enable extra metadata to be stored w/ the key, and 3) differs from the public key input box (hex vs. Base58Check).

Only the third item really needs work before we suggest the signature format for broader use. The first two is on us to make sure it's all more friendly to use in the DV Playground.

Thanks a ton for all the hard work @harlantwood.

+1 from me to merge.

harlantwood:
> 3. The key format is strange because it ... 3) differs from the public key input box (hex vs. Base58Check).

Yeah, this one was weird, so I pushed a fix.

harlantwood:
> There is no way to put in some signed JSON-LD and see if the signature checks out. The Koblitz stuff only checks to see if the displayed information is valid (which is fine, but the DV Playground should probably just have a signature checking mechanism).

I'm not sure if it's quite what you mean, but if you paste in a different (valid) public key ( eg try `1duznTM2aFTdzUnvGx7oBHjWubRpE2EBX`) you will see that the signature (created by signing with an unrelated private key) fails to validate using this public key.  However, if you then paste in the private key that matches this new public key (`L1CwqSsUaGh7NXPB8BfotjdtkzTSWrsRmx1PrtfvK3ZwkgMQRphG`), a new signature is generated, and this one is validated by the matching public key.

I clarified the verbiage a bit in the UI:

- Bitcoin (ECDSA Koblitz) Private Key for Signing
- Bitcoin (ECDSA Koblitz) Public Key for Verification

The latest code is at http://thawing-inlet-70039.herokuapp.com/playground/

harlantwood:
@msporny can we merge this?  Any feedback from others?   @dlongley  @davidlehn @ChristopherA

dlongley:
@harlantwood,

Merged to avoid any further hold up. We can make any additional adjustments as needed via new PRs.

msporny:
Thanks for merging, @dlongley. Sorry, @harlantwood we never meant this PR to hang out there that long.

> I'm not sure if it's quite what you mean

I was raising a general concern wrt. digital signatures and the JSON-LD playground, not with anything you did. What I meant was that "A person can't show up with a digitally signed message and just dump it into the JSON-LD input box and then know whether or not the signature is valid". We have no generalized signature verification mechanism. From what I gather, anything dumped into the JSON-LD input box is then re-signed with the value in "Bitcoin (ECDSA Koblitz) Private Key for Signing" and then checked against the public key in "Bitcoin (ECDSA Koblitz) Public Key for Verification".

harlantwood:
Ah makes sense.  NP, thanks for feedback and merging.

Issue #426, opened 2016-09-20
Proposal: Default content in term definition


gkellogg:
There are a number of different issues regarding the need to define content for a term which is expanded in node values of that term, creating term-specific contexts, or supporting scoped contexts for framing. I think that all of these may be satisfied by specifying default content which is added to the node values of that term.
- # 247 **Scoped Contexts** – This feature seeks to allow `@context` to be used within a term definition, which would allow for additional term definitions to come into affect for values of that term.
- # 262 **JSON key specific default namespaces for values** – This proposal seeks to allow the definition of `@vocab` within a term definition, which would change the way simple keys and types are resolved for values of that term. (Clearly, this could be implemented using #247).
- # 315 **Framing should use embedded contexts for compaction** – This proposal suggested allowing a frame to use different `@context` at different levels, which would be preserved in the framed document. If #247 were supported, this would have the same effect.
- # 369 **Term-level `@base/@vocab`** – Pretty much the same as #262 and could be implemented by #247.
- # 415 **Scoping context based on the `@type`** – describes scoping contexts based on `@type`. This is similar to #247, but rather than basing it on node values of a term, it is automatically applied to node definitions having a specific `@type`. It's not clear what existing mechanism this might hook off of, and IMHO, this is better supported via #247 (or similar).
- We've also seen a general mis-understanding of how `@context` works, with the supposed capability to generate triples based on things in the `@context` (see [Possible to generate implied triples from `@context`?](https://lists.w3.org/Archives/Public/public-linked-json/2015Mar/0038.html), for one).

It strikes me that these could all be supported by a single mechanism to add default content to a node values of a term. Consider, for example:

``` json
{
  "@context": {
    "schema": "http://schema.org/",
    "foaf": "http://xmlns.com/foaf/0.1/",
    "bibo": "http://purl.org/ontology/bibo/",
    "@vocab": "schema",
    "author": {
      "@id": "schema:author",
      "@type": "@id",
      "@content": {
        "@context": {
          "@vocab": "foaf",
          "homepage": {"@id": "foaf:homepage", "@type": "@id"}
        },
        "@type": ["foaf:Person", "schema:Person"]
      }
    }
  },
  "@id": "http://www.w3.org/TR/json-ld/",
  "@type": "bibo:Document",
  "author": {
    "name": "Manu Sporny",
    "homepage": "http://manu.sporny.org/"
  }
}
```

[...]

PR #445, opened 2017-01-02
Add spec text for scoped contexts

### LINKED ISSUE #247: Scoped contexts
While designing contexts to make writing and processing JSON-LD easier there is an issue of polluting the global context namespace with term definitions. It may be beneficial to support scoped contexts or similar functionality. This could make an "enum" pattern much easier to handle.

For instance, if you have a property `http://example.com/foo` with possible values `http://example.com/Bar` and `http://example.com/Baz` you currently could use JSON-LD such as:

``` js
{
  "@context": {
    "ex": "http://example.com/",
    "foo": {
      "@id": "ex:foo",
      "@type": "@vocab"
    },
    "Bar": "ex:Bar",
    "Baz": "ex:Baz"
  },
  "foo": "Bar"
}
```

However, this has now put `Bar` and `Baz` in the "global" namespace when perhaps they are only used as values for `foo`.

A possible scoped context structure could limit how terms are resolved based on current processing state:

``` js
{
  "@context": {
    "ex": "http://example.com/",
    "foo": {
      "@id": "ex:foo",
      "@type": "@vocab"
      "@context": {
        "Bar": "ex:Bar",
        "Baz": "ex:Baz"
      }
    }
  },
  "foo": "Bar"
}
```

The above would limit the `Bar` and `Baz` expansion to only happen when used as a value for the `foo` property. In addition, other properties, such as `bif` could redefine `Bar` and `Baz` to other local values without conflicting with the terms names that `foo` uses.

This feature wouldn't stop anyone from using data however they want, it would just add clarity to contexts without referring to a fully defined schema. Data still needs to be validated by some other means. There is a potential issue for confusion if you have multiple terms with similar names, but I suspect the ease of use benefits would out weight the problems. The alternate less easy to use method is to define top-level terms such as `FooBar` and `BifBar` to avoid conflicts.

Current examples of where this might be of use:

https://github.com/web-payments/payswarm.com/blob/master/contexts/payswarm-v1.jsonld

PaySwarm, as of early 2013, uses a credit card schema and has a context with `cardBrand` with some common brand terms defined such as `Visa`, `MasterCard`, and others. Those values are not used elsewhere. Similar with `bankAccountType`, `currency`, and various other "enum" style properties.

https://developers.google.com/gmail/schemas/reference/rsvp-action

Google, as of May 2013, started using JSON-LD for email actions. In their new RsvpAction type they have an `attendance` property that can take on `http://schema.org/RsvpAttendance/Yes|No|Maybe`. Such properties would be a prime candidate for a scoped context since it's likely the schema used restricts the valid value to those three URIs.

### LINKED ISSUE #262: Feature proposal: JSON key specific default namespaces for values
It would be helpful to be able to define separate default namespaces (@vocab) for the values based on the JSON-key. This would make it easier to hide URIs or namespaces from the data.

For example, when describing a Person, the JSON-LD could look like following pseudo JSON-LD (see "professionalTitle"):

```
{
  "@context": {
    "name": "http://xmlns.com/foaf/0.1/name",
    "homepage": {
      "@id": "http://xmlns.com/foaf/0.1/homepage",
      "@type": "@id"
    },
    "professionalTitle": {
      "@id":"http://example.org/title",
      "@type":"@id",
      "@vocab":"http://example.org/my-titles/"
    }
  },
  "name": "Manu Sporny",
  "homepage": "http://manu.sporny.org/",
  "professionalTitle": "team-leader"
}
```

...instead of the current way where we need to define a namespace, such as "titles:team-leader" or using the full URI.

### LINKED ISSUE #369: JSON-LD 1.1 Feature Request: Term-level @base / @vocab
In certain cases, it would be helpful to have the ability to specify term-level @vocab or @base in the @context. For instance:

``` json
{
  "@context": {
    "as": "urn:example:",
    "foo": {
      "@id": "as:foo",
      "@type": "@vocab",
      "@vocab": "urn:legacy:"
    }
  },
  "foo": "bar"
}
```

When processed, "foo": "bar" expands to "urn:example:foo": "urn:legacy:bar"

The motivating case for this is the migration of Activity Streams 1.0 to 2.0. In AS1, certain properties (verb and objectType) are allowed to be either simple tokens OR fully qualified IRIs. With the simple token, the values do not expand out to IRIs. With the current approach, these would end up expanding to whatever the in-scope Base URL is, which is problematic. What we need is the ability to reliably handle arbitrary token values in a predictable way without requiring that each value be predefined in a context or requiring modifications to the original JSON. The ability to declare a term-level @vocab or @base URI would provide exactly what is needed to address the issue.

gkellogg:
to syntax and API documents along with compaction and expansion tests.

Fixes #247. Fixes #262. Fixes #369.

### DIFF
--- a/spec/latest/json-ld-api/index.html
+++ b/spec/latest/json-ld-api/index.html
@@ -877,7 +877,7 @@
         <li>If <em>value</em> contains the key <code>@type</code>:
           <ol class="algorithm">
             <li>Initialize <em>type</em> to the value associated with the
-              <code>@type</code> key, which MUST be a <a>string</a>. Otherwise, an
+              <code>@type</code> key, which must be a <a>string</a>. Otherwise, an
               <a data-link-for="JsonLdErrorCode">invalid type mapping</a>
               error has been detected and processing is aborted.</li>
             <li>Set <em>type</em> to the result of using the
@@ -976,7 +976,7 @@
         <li>If <em>value</em> contains the key <code>@container</code>:
           <ol class="algorithm">
             <li>Initialize <em>container</em> to the value associated with the
-              <code>@container</code> key, which MUST be either
+              <code>@container</code> key, which must be either
               <code>@list</code>, <code>@set</code>, <code>@index</code>,
               or <code>@language</code>. Otherwise, an
               <a data-link-for="JsonLdErrorCode">invalid container mapping</a>
@@ -992,7 +992,7 @@
             <li>Invoke the <a href="#context-processing-algorithm">Context Processing algorithm</a>
               using the <a>active context</a> and <em>context</em> as <a>local context</a>.
               If any error is detected, an
-              <a data-link-for="JsonLdErrorCode">invalid scoped context</a> errors
+              <a data-link-for="JsonLdErrorCode">invalid scoped context</a> error
               has been detected and processing is aborted.</li>
             <li>Set the <a>local context</a> of <em>definition</em> to <em>context</em>.</li>
           </ol>
@@ -1001,7 +1001,7 @@
           does not contain the key <code>@type</code>:
           <ol class="algorithm">
             <li>Initialize <em>language</em> to the value associated with the
-              <code>@language</code> key, which MUST be either <code>null</code>
+              <code>@language</code> key, which must be either <code>null</code>
               or a <a>string</a>. Otherwise, an
               <a data-link-for="JsonLdErrorCode">invalid language mapping</a>
               error has been detected and processing is aborted.</li>
@@ -1010,6 +1010,11 @@
               of <em>definition</em> to <em>language</em>.</li>
           </ol>
         </li>
+        <li>If the value contains any key other than <code>@id</code>,
+          <code>@reverse</code>, <code>@container</code>,
+          <code>@context</code>, or <code>@type</code>, an
+          <a data-link-for="JsonLdErrorCode">invalid term definition</a> error has
+          been detected and processing is aborted.</li>
         <li>Set the <a>term definition</a> of <em>term</em> in
           <a>active context</a> to <em>definition</em> and set the value
           associated with <em>defined</em>'s key <em>term</em> to
--- a/spec/latest/json-ld-api/index.html
+++ b/spec/latest/json-ld-api/index.html
@@ -620,7 +620,8 @@
       the <a>vocabulary mapping</a> and the <a>default language</a>. Each
       <a>term definition</a> consists of an <dfn data-lt="IRI mappings">IRI mapping</dfn>, a boolean
       flag <dfn data-lt="reverse properties">reverse property</dfn>, an optional <dfn data-lt="type mappings">type mapping</dfn>
-      or <dfn data-lt="language mappings">language mapping</dfn>, and an optional  <dfn data-lt="container mappings">container mapping</dfn>.
+      or <dfn data-lt="language mappings">language mapping</dfn>,
+      an optional context, and an optional <dfn data-lt="container mappings">container mapping</dfn>.
       A <a>term definition</a> can not only be used to map a <a>term</a>
       to an IRI, but also to map a <a>term</a> to a <a>keyword</a>,
       in which case it is referred to as a <dfn data-lt="keyword aliases">keyword alias</dfn>.</p>
@@ -876,7 +877,7 @@
         <li>If <em>value</em> contains the key <code>@type</code>:
           <ol class="algorithm">
             <li>Initialize <em>type</em> to the value associated with the
-              <code>@type</code> key, which must be a <a>string</a>. Otherwise, an
+              <code>@type</code> key, which MUST be a <a>string</a>. Otherwise, an
               <a data-link-for="JsonLdErrorCode">invalid type mapping</a>
               error has been detected and processing is aborted.</li>
             <li>Set <em>type</em> to the result of using the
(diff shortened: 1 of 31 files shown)


gkellogg:
This ended up being a pretty small change to Expansion and Compaction algorithms. Solves three different open issues. Note that #426 would have addressed this too, but it seems to be lacking support.

spec/latest/json-ld-api/index.html
```diff
@@ -876,7 +877,7 @@
         <li>If <em>value</em> contains the key <code>@type</code>:
           <ol class="algorithm">
             <li>Initialize <em>type</em> to the value associated with the
-              <code>@type</code> key, which must be a <a>string</a>. Otherwise, an
+              <code>@type</code> key, which MUST be a <a>string</a>. Otherwise, an
```
lanthaler: This is a property of a conformant document, not a requirement of the algorithm. So shouldn't this remain a lowercase must?
(same elsewhere)

spec/latest/json-ld-api/index.html
```diff
@@ -984,11 +985,23 @@
...
           </ol>
         </li>
+        <li>If <em>value</em> contains the key <code>@context</code>:
+          <ol class="algorithm">
+            <li>Initialize <em>context</em> to the value associated with the
+              <code>@context</code> key, which is treated as a <a>local context</a>.</li>
+            <li>Invoke the <a href="#context-processing-algorithm">Context Processing algorithm</a>
+              using the <a>active context</a> and <em>context</em> as <a>local context</a>.
+              If any error is detected, an
+              <a data-link-for="JsonLdErrorCode">invalid scoped context</a> errors
```
lanthaler: errors -> error

lanthaler: Do we really want swallow the error though? Or should we start supporting nested errors?

spec/latest/json-ld-api/index.html
gkellogg: Debatable, but that seems to be were we came down in the 1.0 document, so we'll use "must".

spec/latest/json-ld-api/index.html
gkellogg: IMO, we go overboard on errors as it is. This tells the author that the problem is with a scoped context, rather than with the context in general, so it seems more useful this way to me.

gkellogg:
> What concerns me is that a conformant JSON-LD 1.0 processor would silently process a document with scoped contexts by simply ignoring them ...

I believe `@context` is still illegal for a JSON-LD 1.0 processor, so if run using processing mode "json-ld-1.0" an error should be generated. For it to be accepted, processing mode must be "json-ld-1.1". The spec says that "json-ld-1.1" is the default processing mode for version 1.1.

From 1.0:

> Initialize container to the value associated with the `@container` key, which must be either `@list`, `@set`, `@index`, or `@language`. Otherwise, an `invalid container mapping` error has been detected and processing is aborted.

gkellogg:
Oops, never mind, wrong change; let me check more; we were pretty careful for looking at such things.

gkellogg:
Looks like you're right. Create Term Definition doesn't look for a constrained list of term definition properties, which it should have. Perhaps we could consider an errata for 1.0 to error on "invalid term definition" if it contains anything other than `@id`, `@reverse`, `@container`, or `@language`, or `@type`. Without this then we could certainly run into a situation where a 1.0 processor sees a 1.1 document and does something different.

Do you have a suggestion for how to handle this? We need to be able to do new things for 1.1.

gkellogg:
@lanthaler thanks for your review! I addressed your specific issues, and added wording about giving an error on unexpected keys in a term definition. Something like that could be considered for a 1.0 errata, but would need to investigate how to make that happen.

Probably best to wait until we're closer to finishing 1.1 in case any other things to be addressed in 1.0 come up.

dlongley:
+1 to merging this. If we can leave an issue open to try and address the 1.0 vs. 1.1 issue in a better way that would be a good thing.

lanthaler:
Term mappings require the value to be null, a string or an object. So we could trigger an issue by a value that is a number (version), a boolean (flag), or an array (list of required features). Something like

```
{
  "@context": {
    "@version": 2,        <---------------------------- triggers an error
    "name": "http://schema.org/name",
    "interest": {
      "@id": "http://xmlns.com/foaf/0.1/interest",
      "@context": {
        "@vocab": "http://xmlns.com/foaf/0.1/"
      }
    }
  },
  ...
}
```

or

```
{
  "@context": {
    "@mustunderstand": [ "nestedcontexts" ],        <--- triggers an error
    "name": "http://schema.org/name",
    "interest": {
      "@id": "http://xmlns.com/foaf/0.1/interest",
      "@context": {
        "@vocab": "http://xmlns.com/foaf/0.1/"
      }
    }
  },
  ...
}
```

gkellogg:
Of there, the `"@version": 2` (better `1.1` as the spec version is titled) is least objectionable, but I'd rather not need to resort to this. Most other things that are being considered for 1.1 will be caught, just not this specific feature. I'd rather handle it with an errata to 1.0.

That said, we might want to have a specific version announcement feature in the context. Perhaps setting the processingMode directly, in which case:

```json
{
  "@context": {
    "@processingMode": "json-ld-1.1",
    ...
  }
}
```

would be a better way to do this, and be consistent with our description of `processingMode` JsonLdOptions value, which could also be an alternative way of specifying the processing model.

The JSON-LD Framing 1.1 document specifies using "json-ld-framing-1.1-expand-frame" as the processing mode to be passed to the Expansion algorithm when processing the JSON-LD Frame (this should probably change to "json-ld-1.1-frame-expansion", or similar, once framing is rolled into the API document). We could consider requiring that frames include this `processingMode` setting to allow for such features, rather than make it implicit.

gkellogg:
@lanthaler If you think your review concerns are satisfied, can you complete the review?

I'm going to add an issue on adding `@version` to the context for enabling such features, although I'm not sold on it myself.

lanthaler:
Sounds good

PR #449, opened 2017-01-08
Issue 430 containers

### LINKED ISSUE #430: Interpreting keys in a @set as @ids or triple objects (RDF)
I've come across this issue several times while trying to interpret raw JSON as JSON-LD and haven't found any practical solution to it.
### Case 1 (index key as `@id`)

Let's say I have a JSON structure like this (minus the context):

```
{
  "@context": {
    "ex": "http://example.org/",
    "properties": { "@id": "ex:properties", "@container": "@set" }
  },
  "properties": {
    "foo" : { },
    "bar" : { }
  }
}
```

In this structure, `foo` and `bar` are identifiers for the sub-objects. So I declare `"properties": { "@container": "@index" }` in my context. However, these IDs are also relevant in RDF and I'd like to keep them while deserializing to RDF. Here is the result I'd like to see (either as blank nodes or as relative URIs, depending on the application):

```
_:b0 ex:property _:foo
_:b0 ex:property _:bar
```

I believe I can't do that with the current JSON-LD specification.
### Case 2 (index key as triple object)

Similarly, I often need index keys to be turned into triple objects (often in a type statement). The [Vehicle Signal Specification](https://github.com/GENIVI/vehicle_signal_specification/) (VSS) is a good example. A VSS object is supposed to describe all available signals in a vehicle in a tree structure where branches are vehicle parts. Here is the first level of a typical VSS object:

```
{
  "children": {
    "Body": {},
    "ADAS": {},
    "Chassis": {}
  }
}
```

If I had a way to declare that each key should be the object of a type statement, I would then just have to deserialize the VSS object using a context like `"@context": { "@vocab": "http://genivi.org/ns/vss#" }` to get the following triples:

```
_:b0 vss:children [ a vss:Body ],
_:b0 vss:children [ a vss:ADAS ],
_:b0 vss:children [ a vss:Chassis ]
```
### Proposal

I believe solving this issue would only require a few declarations in the context (I haven't thought in details about the added complexity on the RDF deserialization algorithm, though).

In the first case, the simplest solution would be to allow `"@container": "@id"` (blank node IDs or IRIs? Needs to be defined). In the second case. allowing `"@container": "@type"` would do most of the job since, from my personal experience, this is the case I've encountered the most.

However, if on wants to allow arbitrary properties as predicate, one could allow any IRI in the container definition, like `"@container": "https://schema.org/productID"`. I know there are already a few issues associated to `@container` (at least #246 and #397) and integrating them all would make things pretty complex. But considering even the first case in JSON-LD 1.1 would be of great help!

gkellogg:
This PR adds support for id-maps and type-maps.
Fixes #430.

Note that I've added some simple change highlighting to make it easier to see changes since 1.0.

For the Syntax doc this includes _Node Identifier Indexing_ and _Node Type_ Indexing sections in addition to the grammar. For the API document, it includes minor changes to _Create Term Definition_, _IRI Compaction_, _Expansion Algorithm_, and _Compaction Algorithm_.

There are also 9 new tests, and one error test marked `json-ld-1.0` only.

### DIFF
--- a/spec/latest/common/terms.html
+++ b/spec/latest/common/terms.html
@@ -110,7 +110,14 @@
       <a>node object</a> or a <a>value object</a> in the input.</dd>
     <dt><dfn data-lt="graph names">graph name</dfn></dt><dd>
       The <a>IRI</a> identifying a <a>named graph</a>.</dd>
-    <dt><dfn data-lt="index maps">index map</dfn></dt><dd>
+    <dt class="changed"><dfn data-lt="id maps">id map</dfn></dt><dd class="changed">
+      An <a>id map</a> is a <a>JSON object</a> value of a <a>term</a> defined with
+      <code>@container</code> set to <code>@id</code>, who's keys are
+      interpreted as <a>IRIs</a> representing the <code>@id</code>
+      of the associated <a>node object</a>; value MUST be a <a>node object</a>.
+      If the value contains a property expanding to <code>@id</code>, it's value MUST
+      be equivalent to the referencing key.</dd>
+     <dt><dfn data-lt="index maps">index map</dfn></dt><dd>
       An <a>index map</a> is a <a>JSON object</a> value of a <a>term</a> defined with
       <code>@container</code> set to <code>@index</code>, who's values MUST be any of the following types:
         <a>string</a>,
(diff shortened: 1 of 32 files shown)


gkellogg:
Note that the formatted HTML versions listed above now are inclusive of changes from PR #451.

PR #451, opened 2017-01-10
Issue 246 nesting

### LINKED ISSUE #246: Ignoring semantically meaningless nesting
Thanks for the great work with JSON-LD! However, when trying to use JSON-LD for to present data in the company I'm working in, I noticed the following missing feature:

FEATURE PROPOSAL: ABILITY TO DEFINE ANY KEY AS AN INDEX KEY

In addition to JSON-LD's existing index container structure, I propose that any key under a JSON-LD node could be defined as a index key.

This would help clustering data under a node into coder friendly logical groups without messing up the Linked Data interpretation with e.g. blank nodes. I encountered the need for this feature at our company where our problem is that the amount of attributes a single JSON-LD node can have can potentially be quite many, say, tens or hundreds of attributes.

As far as I know, this can not be currently done with JSON-LD without 1) ending up with blank nodes or  2) the need to create a deeper JSON structure by using a separate index term (using "@container":"@index") which then contains the data underneath.

In addition, if a single key could be defined as a index term, this would make it more flexible to attach the JSON-LD Linked Data interpretation to even a wider amount of existing JSON data, without having to change the structure of such data (and without ending up with e.g. lots of blank nodes).

DEFINING AN INDIVIDUAL INDEX KEY IN @CONTEXT

The "@context" definition could be done e.g. using the existing reserved keyword "@index" in the following way:

```
"indexkey":"@index"
```

which should be interpreted in the following way: 1) the "indexkey" is an index key and should be skipped when traversing the JSON tree while doing the JSON-LD to RDF interpretation, 2) any data directly under the "indexkey" should be interpreted as data directly attached to the node of the indexkey (same RDF subject).

EXAMPLE

To give a full example, in the following a single key "labels" is defined as an index index key to help grouping the data into coder friendly logical groups without messing up the Linked Data interpretation):

``` js
{
  "@context": {
     "labels":"@index",
     "main_label":"http://example.org/my-schema#main_label",
     "other_label":"http://example.org/my-schema#other_label",
     "homepage":{ "@id":"http://example.org/my-schema#homepage", "@type":"@id"}
  },
  "@id":"http://example.org/myresource",
  "homepage": "http://example.org",
  "labels": {
     "main_label": "This is the main label for my resource",
     "other_label": "This is the other label"
  }
}
```

This example JSON-LD should generate the following RDF triplets:

```
<http://example.org/myresource> <http://example.org/my-schema#homepage> <http://example.org>. 
<http://example.org/myresource> <http://example.org/my-schema#main_label> "This is the main label for my resource".
<http://example.org/myresource> <http://example.org/my-schema#other_label> "This is the other label".
```

gkellogg:
This introduces the `@nest` keyword, used for folding contained properties to the enclosing node object, and allows for `@container: @nest` in a term definition, to introduce such nesting when compacting.

Fixes #246.

This PR should not be merged before #449, as it builds on changes made there.

There are several new expand, compact and error tests for this feature.

### DIFF
--- a/spec/latest/common/terms.html
+++ b/spec/latest/common/terms.html
@@ -180,6 +180,10 @@
       specified via the <code>@context</code> <a>keyword</a>.</dd>
     <dt><dfn data-lt="named graphs">named graph</dfn></dt><dd>
       A <a>linked data graph</a> that is identified by an <a>IRI</a> or <a>blank node</a>
+    <dt><dfn data-lt="nested properties">nested property</dfn></dt><dd>
+      A <a>nested property</a> is a <a>property</a> which is contained within an object referenced by
+      a semantically meaningless <em>nesting property</em>.
+    </dd>
       (the <a>graph name</a>) and a <a>graph</a>.</dd>
     <dt><dfn data-lt="nodes">node</dfn></dt><dd>
       Every <a>node</a> is an <a>IRI</a>, a <a>blank node</a>,
--- a/spec/latest/json-ld-api/index.html
+++ b/spec/latest/json-ld-api/index.html
@@ -987,8 +987,8 @@
             <li>Initialize <em>container</em> to the value associated with the
               <code>@container</code> key, which must be either
               <code>@list</code>, <code>@set</code>, <code>@index</code>,
-              <span class="changed"><code>@id</code>, <code>@type</code></span>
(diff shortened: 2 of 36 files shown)


gkellogg:
Obsoleted by PR #453.

PR #453, opened 2017-01-12
This adds a `@nest` member to term definitions used for _transparent properties_

### LINKED ISSUE #246: Ignoring semantically meaningless nesting

gkellogg:
This introduces the `@nest` keyword, used for folding contained properties to the enclosing node object, and allows for `@nest` in a term definition, to introduce such nesting when compacting.

Fixes #246. This also is an alternative to PR #451 and is (mostly) independent of #449).

In contrast to #451, having a parallel `@nest` term definition property allows this to be used in conjunction with `@container` (but not for `@reverse` compaction).

There are several new expand, compact and error tests for this feature.

### DIFF
--- a/spec/latest/json-ld-api/index.html
+++ b/spec/latest/json-ld-api/index.html
@@ -4102,6 +4102,7 @@
             "cyclic IRI mapping",
             "invalid @id value",
             "invalid @index value",
+            "invalid @nest value",
             "invalid @reverse value",
             "invalid base IRI",
             "invalid container mapping",
@@ -4112,7 +4113,6 @@
             "invalid language mapping",
             "invalid language-tagged string",
             "invalid language-tagged value",
-            "invalid @nest value",
             "invalid local context",
             "invalid remote context",
             "invalid reverse property",
@@ -4154,6 +4154,8 @@
         <dt><dfn>invalid @index value</dfn></dt>
         <dd>An <code>@index</code> member was encountered whose value was
           not a <a>string</a>.</dd>
+        <dt class="changed"><dfn>invalid @nest value</dfn></dt>
+        <dd class="changed">An invalid value for <code>@nest</code> has been found.</dd>
         <dt><dfn>invalid @reverse value</dfn></dt>
         <dd>An invalid value for an <code>@reverse</code> member has been detected,
           i.e., the value was not a <a>JSON object</a>.</dd>
@@ -4189,8 +4191,6 @@
           associated language tag was detected.</dd>
         <dt><dfn>invalid local context</dfn></dt>
         <dd>In invalid <a>local context</a> was detected.</dd>
-        <dt class="changed"><dfn>invalid @nest value</dfn></dt>
-        <dd class="changed">An invalid value for <code>@nest</code> has been found.</dd>
         <dt><dfn>invalid remote context</dfn></dt>
         <dd>No valid context document has been found for a referenced,
          remote context.</dd>
--- a/spec/latest/json-ld-api/index.html
+++ b/spec/latest/json-ld-api/index.html
@@ -903,7 +903,7 @@
         </li>
         <li>If <em>value</em> contains the key <code>@reverse</code>:
           <ol class="algorithm">
-            <li>If <em>value</em> contains an <code>@id</code>, member, an
+            <li>If <em>value</em> contains <code>@id</code> or <code>@nest</code>, members, an
               <a data-link-for="JsonLdErrorCode">invalid reverse property</a>
               error has been detected and processing is aborted.</li>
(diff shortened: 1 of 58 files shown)


spec/latest/json-ld-api/index.html
```diff
@@ -4058,6 +4112,7 @@
             "invalid language mapping",
             "invalid language-tagged string",
             "invalid language-tagged value",
+            "invalid @nest value",
```
davidlehn: move up sorted with other @ values

spec/latest/json-ld-api/index.html
```diff
@@ -4134,6 +4189,8 @@
           associated language tag was detected.</dd>
         <dt><dfn>invalid local context</dfn></dt>
         <dd>In invalid <a>local context</a> was detected.</dd>
+        <dt class="changed"><dfn>invalid @nest value</dfn></dt>
```
davidlehn: move up sorted with other @ values

spec/latest/json-ld-api/index.html
gkellogg: Good call, the `@` was added later. Should be good now.

gkellogg:
@dbooth-boston Flattening is an entirely different concept, which takes nested node definitions and flattens them so that nodes only reference other nodes, rather than include them.

@dlongley suggested `@preserve`, but that has a different use in framing.

However, the use case you may be looking for could be handled using scoped contexts, there interpretation of "location" is different for "file" vs "vehicle". This was the subject of a recently merged change. But, that does imply that both "file" and "vehicle" refer to different entities, rather than are each properties of the containing node, which is more consistent with what I think you want.

As for a `@nest` alternative, I'm open to suggestions.

dbooth-boston:
I was thinking of _.flatten from http://underscorejs.org/#flatten -- not the JSON-LD concept of flattening. I should not have used that word.  Sorry for the confusion!

Yes, the use case I was mentioning would be the ability to distinguish between properties (of the same name) by their object ancestors, as illustrated above.  This came up a couple of years ago when we were originally trying to see if FHIR/JSON could be adapted to be JSON-LD.  We concluded that it could not without either changing the existing property names or inserting additional @contexts into the instance data -- neither of which would be palatable to the vanilla FHIR/JSON folks.

PR #464, opened 2017-02-07
Added failing test for digitalbazaar/jsonld.js#167

jmatsushita:


test-suite/tests/compact-manifest.jsonld
```diff
@@ -586,6 +586,14 @@
...
       "expect": "compact-0072-out.jsonld"
     }, {
+      "@id": "#t0073",
+      "@type": ["jld:PositiveEvaluationTest", "jld:CompactTest"],
+      "name": "Mapped @id and @type",
+      "purpose": "Ensure that compaction works with mapped @id and @type",
+      "input": "compact-0073-in.jsonld",
+      "context": "compact-0073-context.jsonld",
+      "expect": "compact-0073-out.jsonld"
+      }, {
```
gkellogg: For this to work consistently, you need to add "option": {"base": "http://example/base/"} or similar to the test definition.

gkellogg:
Also, the version of the test suite you're using doesn't seem to be the current master branch, which includes tests after 0072.

gkellogg:
@jmatsushita @dlongley I someone want's to fix the PR, we can incorporate the test, otherwise close without prejudice?

jmatsushita:
Hi there, bumped into this bug again in jsonld.js digitalbazaar/jsonld.js#167 and thought I'd update this PR. 

> includes tests after 0072.

Yes I did see this but got confused with the numbering (which is also lettering now which I don't if its semantic somehow). Can you tell me which number to use?

> This seems like a reasonable test, although by default, @type has type @vocab, not @id, but that may be the point of your test.

It wasn't actually so I've simplified the test which still yields an error in jsonld.js 0.4.11 digitalbazaar/jsonld.js#167

jmatsushita:
Ug. I made the history ugly. Happy to close this PR and reopen a new one.

gkellogg:
> Yes I did see this but got confused with the numbering (which is also lettering now which I don't if its semantic somehow). Can you tell me which number to use?

Identifiers don't really mean anything, I used some arbitrarily based on the feature which was being added, specifically so we didn't get into naming conflicts, As long as your identifier generally fits in the scheme and does not conflict with an existing identifier, you should be good to go.

gkellogg:
Yes, please create a fresh PR based on the current "master" branch, which should only add the new test files and update the manifest.

PR #472, opened 2017-03-30
Add CSP: upgrade-insecure-requests

BigBlueHat:
Fixes the Mixed Content mess...for real.

I have this in place on the [LevelGraph Playground](http://wileylabs.github.io/levelgraph-playground) and it works.

Cheers!
🎩

### DIFF
--- a/playground/index.html
+++ b/playground/index.html
@@ -14,7 +14,7 @@
     <!-- Meta Tags -->
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <meta http-equiv="Content-Type" content="text/html;charset=utf-8" />
-
+    <meta http-equiv="Content-Security-Policy" content="upgrade-insecure-requests">
 
     <!-- Style Sheets -->
     <link rel="stylesheet" type="text/css" href="../static/css/bootstrap/bootstrap.css">


gkellogg:
Looks like this is causing the playground to not load, I'm going to back it out.

davidlehn:
http://caniuse.com/#search=upgrade-insecure-requests
This is not universally supported.  I just tested in Edge by adding "http://schema.org" to the `@context` in that above link.  It output a console "Access is deined." error.  So some other solution is still needed if the desire is to just blindly switch protocols.

I think sites should warn users if they are doing this.  Or maybe make it an option somehow.  In practice it's probably true that most sites serve the same data on https and http.  But that's not a requirement.  It's also very possible that a server for a http URL might not support https.  What happens in that case?

gkellogg:
Site doesn't run HTTPS quite yet, once it does, we can re-apply this.

cc/ @davidlehn

BigBlueHat:
Yeah. Sorry about this mess. It's not of my making of course. :wink: But obviously these attempts to fix the mess the browser are making by "aggressively deprecating HTTP" with these sort of sanctioned "hacks" aren't really going to solve for all the scenarios possible--especially in a playground such as this.

I'd also forgotten (apologies) that the json-ld.org site wasn't using HTTPS yet--whereas the [LevelGraph Playground](http://wileylabs.github.io/levelgraph-playground) is because of `*.github.io` URLs all being HTTPS now.

Anyhow. Apologies for yet-more-mess trying to solve for these scenarios.

I do hope some leadership in the Linked Data space surfaces that addresses this "rename all the things" situation...

Sorry again,
🎩

Issue #446, opened 2017-01-06
@version announcement for enabling 1.1 features


gkellogg:
In PR #445, @lanthaler had a concern that a 1.0 processor would silently ignore extra properties of an expanded term definition, such as `@context` as specified for #247. As it happens, this is one of the few places the 1.0 algorithms didn't look for extra values to raise an exception.

One way to do this is using the `processingMode` API option, which is defined to be `json-ld-1.1` for the 1.1, or `json-ld-1.0` for the Recommendation. But, in most cases, this can't be specified, and there's no way to determine what version a document is intended for.

One solution is require that a context using 1.1 features specifically announce this using `@version`.

```json
{
  "@context": {
    "@version": 1.1,
    "term-using-1.1-features": {"@context": {...}}
  }
}
```

I would say that either `processingMode` is specified explicitly (keeping the default at `json-ld-1.0`), or that the context contains `@version`. One consideration is that Framing now uses `processingMode: json-ld-1.1-expand-frame` to allow the special features in a frame document, and adding this using `@version` would be problematic.

That said, another solution would be to create an erratum on 1.0 to add better checking for keys in an expanded term definition, and possibly other places, as may prove necessary.

gkellogg:
See @lanthaler's comment in https://github.com/json-ld/json-ld.org/pull/445#pullrequestreview-15008285

akuckartz:
I dislike ‘version‘ properties.

gkellogg:
@akuckartz  Do you have an alternative suggestion for how to ensure that 1.0 implementations don't silently ignore 1.1 features?

gkellogg:
My thought here is to not define a default for `processingMode` in the API. In 1.0, the processingMode defaults to `json-ld-1.0`. In 1.1, I would say that the default comes from the context used for expansion or compaction.

If not specified through the API, Context processing establishes a `version` in the first context that is a _JSON Object_ that could contain `@version`. If it is specified, it **must** be `1.1` (future versions of JSON-LD may establish other numeric values). If version 1.0 has been established (via API or the absence of `@version` in the first encountered JSON Object context), encountering a `@version: 1.1` is an error.

This allows legacy contexts to be used along with 1.1 contexts, as long as the 1.1 context is established first, or the version is set explicitly through the API. 1.0 processors will (should) have implicitly set version to 1.0, and will (should) error when encountering a `@version` key in a context.

The consequence for this when expanding is that the version cannot be absolutely established until after context processing.

[...]

PR #473, opened 2017-04-08
Issue 446 version announcement

gkellogg:
Adds announcement requirement and tests as required in #446.


test-suite/tests/error-p005-context.jsonld
```diff
@@ -0,0 +1,5 @@
+{
+  "@context": {
+    "@version": 1.1
```
dlongley: Do we want this to be a number (float/double) or a string? Expressing version attributes as strings seems to be fairly common practice.

spec/latest/common/terms.html
```diff
@@ -212,6 +212,14 @@
       A <a>prefix</a> is a <a>term</a> that expands to a vocabulary <a>base IRI</a>. It
       is typically used along with a <em>suffix</em> to form a <a>compact IRI</a> to create an IRI
       within a vocabulary.</dd>
+    <dt><dfn>processing mode</dfn></dt><dd>
+      The processing mode defines how a JSON-LD document is processed.
+      By default, all documents are assumed to be conformat with
+      <a data-cite="JSON-LD-20140116">JSON-LD 1.0</a> [[!JSON-LD-20140116]]. By defining
+      a different version using the <code>@version</code> member in a
```
dlongley: I suppose there may be good reasons for decoupling the version from the processing mode, but figured I'd throw out the idea that we could do `@processingMode` instead of `@version` as a means by which to set the preferred processing mode via the `@context` -- rather than inferring it from `@version`.

test-suite/tests/error-p005-context.jsonld
gkellogg: @lanthaler suggested a number, as using a string would be silently ignored by most 1.0 processors; the point is to use something that would cause a 1.0 processor to halt, thus the number.

spec/latest/common/terms.html
gkellogg: That thought occurred to me, but it would be more natural, then, to say `"@processingMode": "json-ld-1.1"`, which would pass a 1.0 processor silently.

An alternative suggested by @lanthaler was to use something like `"@processingMode": ["json-ld-1.1"]`, which would error by a 1.0 processor, would would allow the symmetry with the API mode. Personally, I preferred using the simpler numeric value,  but I could be convinced to go the other way.

test-suite/tests/error-p005-context.jsonld
dlongley: Ah, right, I forgot about that. Bummer.

spec/latest/common/terms.html
dlongley: I don't feel that strongly, so I'm ok with going forward with the numeric proposal for now given its necessity for causing a 1.0 processor to halt.

gkellogg:
Thanks @dlongley: I'll add some additional text to clarify that the value is the JSON number `1.1`.

PR #478, opened 2017-04-09
In expansion, allow a language map to have a `null` value

### LINKED ISSUE #375: Inconsistency between Syntax and API specs
The [JSON-LD Syntax spec](http://www.w3.org/TR/json-ld/#language-maps) indicates that the values of keys in a language map may be `null`. However, the [Expansion Algorithm](http://www.w3.org/TR/json-ld-api/#expansion-algorithm) raises an error if a value is not a string (therefore not allowing `null`). See:

```
7.5.2.2.1) item must be a string, otherwise an invalid language map value error has been detected and processing is aborted.
```

Similarly, the [invalid language map value error definition](http://www.w3.org/TR/json-ld-api/#idl-def-JsonLdErrorCode.invalid-language-map-value) in the API spec doesn't mention allowing `null`.

@lanthaler, @gkellogg, @msporny -- thoughts on the proper resolution? Which spec do you think is incorrect?

I believe we have no tests for using `null` values in a language map -- and if I modify `jsonld.js` to allow `null`, no other tests fail.

Related to https://github.com/digitalbazaar/jsonld.js/issues/76.

gkellogg:
which is allowed in the syntax. Such values are ignored.

Fixes #375.

### DIFF
--- a/spec/latest/json-ld-api/index.html
+++ b/spec/latest/json-ld-api/index.html
@@ -1484,7 +1484,7 @@
                       <em>language value</em>.</li>
                     <li>For each <em>item</em> in <em>language value</em>:
                       <ol class="algorithm">
-                        <li><em>item</em> must be a <a>string</a>,
+                        <li><em>item</em> must be a <a>string</a>
                           <span class="changed">or <code>null</code></span>,
                           otherwise an
                           <a data-link-for="JsonLdErrorCode">invalid language map value</a>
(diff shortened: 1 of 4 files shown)


gkellogg:
@dlongley: note this does not address step 7.4.7, as [you suggested](https://github.com/json-ld/json-ld.org/issues/375#issuecomment-73391180). It didn't seem necessary given the syntax requirement. See [my comment in the issue](https://github.com/json-ld/json-ld.org/issues/375#issuecomment-292818532).

spec/latest/json-ld-api/index.html
```diff
@@ -1485,14 +1485,16 @@
                     <li>For each <em>item</em> in <em>language value</em>:
                       <ol class="algorithm">
                         <li><em>item</em> must be a <a>string</a>,
```
dlongley: Can nix the comma at the end of the line here

PR #479, opened 2017-04-09
Add text clarifying what an "unconnected node" is.

### LINKED ISSUE #418: Section 7 of syntax spec could be made more clear
In [Section 7 "Data Model"](https://www.w3.org/TR/json-ld/#data-model), this statement is made:

> A graph must not contain unconnected nodes, i.e., nodes which are not connected by an edge to any other node.

What this means in practice is that this JSON-LD document:

``` jsonld
{
  "@id": "http://example.org/1"
}
```

Generates no triples (has no data). I think people may come to the conclusion that you can't have disjoint nodes in a graph or other erroneous ideas without some examples to better explain what that statement actually means. Even upon re-reading it myself I was confused about it for a short while.

gkellogg:
Fixes #418.

### DIFF
--- a/spec/latest/json-ld/index.html
+++ b/spec/latest/json-ld/index.html
@@ -3076,7 +3076,21 @@ specified. The full <a>IRI</a> for
     <li>A <a>node</a> having an outgoing edge MUST be an <a>IRI</a> or a
       <a>blank node</a>.</li>
     <li>A <a>graph</a> MUST NOT contain unconnected <a>nodes</a>,
-      i.e., nodes which are not connected by an <a>edge</a> to any other <a>node</a>.</li>
+      i.e., nodes which are not connected by an <a>edge</a> to any other <a>node</a>.
+      <pre class="illegal-example"
+          data-transform="updateExample"
+          title="Illegal Unconnected Node">
+      <!--
+      {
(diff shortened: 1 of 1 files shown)


gkellogg:
@dlongley If this resolution is satisfactory, we can just merge it.

spec/latest/json-ld/index.html
```diff
@@ -3076,7 +3076,21 @@
...
+          data-transform="updateExample"
+          title="Illegal Unconnected Node">
+      <!--
+      {
+        "@id": "http://example.org/1"
+      }
+      -->
+      </pre>
+      <div class="note">
+        This effectivly just bans node objects containing only an <code>@id</code>.
```
dlongley: typo: "effectivly" => "effectively"

dlongley: I can see someone (who does not read the rest of the spec, which is common) getting confused by this wording and also thinking this is illegal:

```js
{
  "foo:bar": {
    "@id": "http://example.org/1"
  }
}
```

As that nested node contains only an `@id`. Perhaps a slight tweak:

`This effectively prohibits unnested, empty node objects and unnested node objects that contain only an <code>@id</code>.`

PR #485, opened 2017-05-04
Issue 463 data model

gkellogg:
This PR updates the description to abstract JSON from the core representation by describing the "JSON-LD internal representation" appropriate for describing the behavior of algorithms. This makes room for other serializations congruent to JSON to use the same algorithms (e.g., YAML, CBOR, ...).


danbri:
Do all file formats eventually define something like an 'infoset' ? Is this broadly equivalent to https://en.wikipedia.org/wiki/XML_Information_Set ?

gkellogg:
Yes, it could be called the "JSON Infoset", although the XML Infoset is more about preserving structural elements. For JSON, all data types have fairly natural native representations within a programming language; it's "JSON Object" which is the main concept to be abstracted.

spec/latest/json-ld-api/index.html
```diff
@@ -504,7 +513,7 @@
     </pre>
 
     <p>Note how in the output above all properties of a <a>node</a> are collected in a
-      single <a>JSON object</a> and how the <a>blank node</a> representing
+      single <a class="changed">dictionary</a> and how the <a>blank node</a> representing
```
dlongley: Here when we say: "Note how in the output above...", we're conflating the output of the algorithm (which is in "internal representation" form) with what's actually shown above, which is the JSON-LD serialization of that internal representation.

It seems like we need some common phraseology when talking about JSON-LD documents vs. internal representation and the examples of inputs/outputs. The examples are all going to be JSON-LD documents -- and so if we just jump into internal representation terminology to talk about them that may be confusing. Similarly, if we talk about the actual outputs of algorithms loosely without distinguishing between whether we're talking about the internal representation or the JSON-LD document that is a result of serializing it, we may cause confusion.

Perhaps we can just sprinkle in something like:

"The example above is the JSON-LD serialization of the output of the flattening algorithm. Note how ... JSON object ..."

Here we'd use JSON object because we're talking about the serialized output. Alternatively, we could go with:

"The example above is the JSON-LD serialization of the output of the flattening algorithm. Note how the algorithm's output ... dictionary ..."

Where we'd use "dictionary" because we're talking about the output itself. I think we may want to add these sorts of phrases where examples collide with talking about algorithm output in the spec.

spec/latest/json-ld-framing/index.html
```diff
@@ -177,7 +177,7 @@
 <p>A JSON-LD document is a representation of a directed graph. A single
   directed graph can have many different serializations, each expressing
   exactly the same information. Developers typically work with trees, represented as
-  <a>JSON objects</a>. While mapping a graph to
+  <a class="changed">dictionaries</a>. While mapping a graph to
```
dlongley: I think we want to say "JSON objects" here.

spec/latest/json-ld/index.html
```diff
@@ -333,10 +334,25 @@
...
+
+    <p>At the surface level, a <a>JSON-LD document</a> is simply
+      <a data-cite="RFC7159">JSON</a>, detailed in [[!RFC7159]].
+      For the purpose of describing the core data structures,
+      this is limited to <a>arrays</a>, <a>dictionaries</a> (the parsed version of a <a>JSON Object</a>),
+      <a>strings</a>, <a>numbers</a>, <a>booleans</a>, and <a>null</a>,
+      <span class="changed">called the <a>JSON-LD internal representation</a>.
+        This allows surface syntaxes other than JSON
+        to be manipulated using the same algorithms, when the syntax maps
+        to equivalent core data structures.</span>.</p>
```
dlongley: Double period

spec/latest/json-ld/index.html
```diff
@@ -3790,6 +3741,9 @@
       <code>@nest</code> mapping. When expanding, the values of a property
       expanding to <code>@nest</code> are treated as if they were contained
       within the enclosing <a>node object</a> directly.</li>
+    <li>The JSON syntax has been abstracted into an <a>internal representation</a>
+      to allow for other seriazation formats that are functionally equivalent
```
dlongley: typo: "seriazation" => "serialization"

Could maybe just say "serializations" instead of "serialization formats" to save a word :)

spec/latest/json-ld-api/index.html
```diff
@@ -448,9 +457,9 @@
       <dfn data-lt="flattened">flattening</dfn> goes a step further to ensure that the shape of the data
       is deterministic. In expanded documents, the properties of a single
       <a>node</a> may be spread across a number of different
-      <a>JSON objects</a>. By flattening a
+      <a class="changed">dictionaries</a>. By flattening a
```
gkellogg: The algorithms operate on the internal representation, and this is a description of the algorithm.

I did give a thought to using "abstract representation" (or "abstract syntax"), but that is not as helpful for discussing a transformation. Danbri's "Infoset" idea is also similar.

The point here, though, is to separate the algorithms from the serialization.

spec/latest/json-ld-api/index.html
gkellogg: I attempt to work the algorithms to deal just with the internal representation, and leave serialization to/from to the API wrapper.

I like the idea of describing the "JSON serialization" of the internal representation to clarify this. It's pretty equivalent to RDF Concepts use of N-Triples, which is not the same as the RDF Abstract Syntax.

I'll work on this some more.

spec/latest/json-ld-framing/index.html
gkellogg: Yes, because we're discussing serializations.

gkellogg:
@dlongley I added an "Example Conventions" section, and noted after appropriate examples that the format is a JSON serialization of the internal form used by the algorithms.

dlongley:
LGTM

PR #487, opened 2017-05-10
Change editors/authors for 1.1.

msporny:
Alternative structure for editors/authors. We will need to override the specref database after we agree to this ordering. The specref changes should be:

* JSON-LD 1.1 API SpecRef Authors: Dave Longley, Gregg Kellogg, Markus Lanthaler, Manu Sporny, Niklas Lindstrom
* JSON-LD 1.1 SpecRef Authors: Manu Sporny, Dave Longley, Gregg Kellogg, Markus Lanthaler, Niklas Lindstrom


spec/latest/json-ld-api/index.html
```diff
@@ -40,33 +40,48 @@
...
-            company: "Kellogg Associates", companyURL: "http://kellogg-assoc.com/" },
-          { name: "Markus Lanthaler", url: "http://www.markus-lanthaler.com/",
-            company: "Graz University of Technology", companyURL: "http://www.tugraz.at/" },
-          { name: "Manu Sporny", url: "http://digitalbazaar.com/",
-            company: "Digital Bazaar", companyURL: "http://digitalbazaar.com/" }
+        { name:       "Dave Longley",
+          url:        "http://digitalbazaar.com/",
+          company:    "Digital Bazaar",
+          companyURL: "http://digitalbazaar.com/",
+          note:       "v1.0" },
```
lanthaler: Why do you note the versions for "authors"? The vast majority of the content didn't change between the versions

spec/latest/json-ld-api/index.html
```diff
@@ -40,33 +40,48 @@
...
       includePermalinks:      true,
       noRecTrack:             false,
-      testSuiteURIkey:       "http://json-ld.org/test-suite/", 
+      testSuiteURIkey:       "http://json-ld.org/test-suite/",
       postProcess:            [internalizeTermListReferences],
 
       // editors, add as many as you like
       // only "name" is required
       editors:  [
-          { name: "Markus Lanthaler", url: "http://www.markus-lanthaler.com/",
```
lanthaler: To be honest, I'm not too happy about these changes given [how much work](https://github.com/json-ld/json-ld.org/graphs/contributors) I put into this

gkellogg: When specs are updated, it's quite common that editors change. Manu's approach preserves editors and authors from 1.0 into 1.1. I believe this was the case when you edited RDF Concepts 1.1.

The fact is, even if there is not a substantial amount of change to the document, there is considerable effort put in by me to make it happen. In many cases, reducing the change is more work than changing things more.

The fact is, we all put a lot of work into 1.0, and I've put a lot of work into the 1.1 effort.

What would you say a fair way to deal with this is?

spec/latest/json-ld-api/index.html
gkellogg: See above and look at the number of commits and PRs post 1.0 to get a sense for the amount of effort put into the 1.1 work, and we're not done quite yet.

spec/latest/json-ld-api/index.html
msporny: @lanthaler We're trying to accomplish a number of things here:

1. Ensure that people that contributed to the spec show up in the SpecRef database properly.
2. Ensure that we are consistent in how we decide who Editors are and who Lead Designers are.

Editors for a spec are the ones that do the editorial work on that version spec. You were at the top for JSON-LD 1.0 API:

https://www.w3.org/TR/json-ld-api/

Gregg is the only one that has really worked on JSON-LD 1.1. While it may be true that a good bit of the text did not change, it's also true that Gregg has done 99%+ of the editorial work for JSON-LD 1.1... and that's what the "Editors" line is for in the spec... the people that did the work /for that version of the spec/.

As far as the "Lead Designers"/Authors... we have that so that people that didn't necessarily edit a great deal, but contributed a large amount in other ways, are properly credited. We need to make sure these people are properly reflected in the SpecRef database.

Dave Longley went through multiple algorithm implementations many times before the JSON-LD specs were made public. He implemented all the algorithms in 4 different languages (C++ initially, Javascript, PHP, and Python) and then documented them in the specs before anyone else picked up the specs and started working on them. He also made a big pass at the end to simplify the explanations of the algorithms. However, when the JSON-LD API is referenced via SpecRef, he doesn't show up in the credits at all because he is listed as an author. Clearly, that is not fair to him.

That's why we have the "Lead Designers"/Authors section. It's to acknowledge all of the other vital work that goes into a spec - thinking, designing, implementing, debugging, etc.

I went through and attempted to weigh all of the work that was done by everyone, both editorial and non-editorial, and generated the lists in such a way that everyone is fairly acknowledged for the work that they have done over time. 

For the JSON-LD API, the SpecRef database will have this order: Dave Longley, Gregg Kellogg, Markus Lanthaler, Manu Sporny, Niklas Lindstrom ... because that order constitutes the total amount of value that each person has contributed to the spec over time.

To put it in different terms: 

* I don't think anyone can make a compelling case that they provided more value than Dave Longley did for the JSON-LD algorithms. So, he goes first.
* While it may have been true that Markus provided the next bit of value on the algorithms for 1.0, Gregg was also a prominent figure during the 1.0 work. That, in addition to Gregg effectively single-handedly taking on JSON-LD 1.1 and moving it forward bumps him up to second place in the list with Markus close behind.

I won't go on to describe Niklas' and my contributions to the API spec as I don't think either of us (or anyone else) are contesting them.

Does this help explain the reasoning behind the order, Markus? If this still doesn't sit well with you, please suggest an alternative as keeping it as-is definitely isn't fair to Dave Longley and I'd argue it wouldn't be fair to  Gregg either.

spec/latest/json-ld-api/index.html
msporny: > Why do you note the versions for "authors"?

So that people are more accurately attributed for the work that they did.

PR #501, opened 2017-05-24
Update IRI Compaction, Compaction, and Remove Preserve logic to cause…

### LINKED ISSUE #496: Framing doesn't compact a property if it is missing from the data (even when default is provided)
@gkellogg sorry to bother you once more, I think this should be my last issue in my (miss)-handling of "@type" that causes me to get unexpected behavior.  

Consider this data:

```
{
  "@context": {
    "schema": "http://schema.org/",
    "name": "schema:name",
    "url": {"@id": "schema:url", "@type": "schema:URL"}
  },
  "name": "Jane Doe"
}
```

Under this frame:

```
{
  "@context": {
    "schema": "http://schema.org/",
    "name": "schema:name",
    "url": { "@id": "schema:url", "@type": "schema:URL"}
  },
  "name": {},
  "url": {"@default": "http://example.com"}
}
```

The output `url` property cannot be compacted and remains as `schema:url`.

On one hand, this makes sense -- this is the same compaction behavior we would get if the input data did contain `url` but with a different type.  Because the element is not present, that's kinda like not knowing it is typed correctly.  

However, it would seem that this is clearly not the desired behavior for the developer, who expects to still be able to reference `url` and get back the default value.  This all works as expected if the context doesn't declare an explicit data type for `url`.  Perhaps a bug? But again I'm likely missing something.  Thanks!

[Playground](http://tinyurl.com/kkcm4zv)

gkellogg:
… values of `@preserve` to be compacted using the referencing term, and for term selection to consider the value of `@preserve`, rather than `@preserve` itself.

Adds a test to ensure appropriate handling by processors.
Fixes #496.

### DIFF
--- a/spec/latest/json-ld-api/index.html
+++ b/spec/latest/json-ld-api/index.html
@@ -2085,6 +2085,18 @@
                 <li>Continue with the next <em>expanded property</em> from <em>element</em>.</li>
               </ol>
             </li>
+            <li class="changed">If <em>expanded property</em> is <code>@preserve</code>
+              then:
+              <ol class="algorithm">
+                <li>Initialize <em>compacted value</em> to the result of using this
+                  algorithm recursively, passing <a>active context</a>,
+                  <a>inverse context</a>, <em>property</em> for
+                  <a>active property</a>, and <em>expanded value</em>
+                  for <em>element</em>.</li>
+                <li>Add <em>expanded value</em> as the value of <code>@preserve</code>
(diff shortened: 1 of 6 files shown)


dlongley:
JFYI -- I do see this and related issue(s) ... just haven't had time to review yet. Will get to it when I can, not trying to leave you hanging, @gkellogg :).

gkellogg:
@dlongley thanks, this is a pretty small update. No rush.

gkellogg:
@dlongley I'll plan on merging this tomorrow, unless you have objections.

PR #503, opened 2017-05-28
@container as array including @set.

### LINKED ISSUE #269: Ability to declare a container as both @index and @set
I have the following indexed data structure that I'd like to preserve during compaction:

```
{
  "dbxref": {
    "db1": ["term1"],
    "db2": ["term2a", "term2b"]
  }
}
```

Declaring a context with @container: @index partly works, but I also want to preserve the array even when it has only a single member:

```
{
  "@context": {
    "@vocab": "http://localhost/context#",
    "dbxref": {
      "@container": "@index"
    }
  },
  "@id": "http://localhost/item1",
  "dbxref": {
    "db1": "term1",
    "db2": ["term2a", "term2b"]
  }
}
```

It would be helpful to have some way to declare a container as both @index and @set.

### LINKED ISSUE #407: Ability to declare a container as both @language and @set
Related to #269 . I would like to use `@language` maps with `@set`, but haven't found a way to do so.

My data is following the [SKOS model](http://www.w3.org/TR/skos-reference/), where each concept can have one preferred label per language, and any number of alternative labels per language. To make the data easy to work with, I'm trying to achieve the following output:

``` json
{
  "@id": "http://data.ub.uio.no/realfagstermer/c012548",
  "prefLabel": {
    "nb": "Abbor",
    "en": "European perch"
  },
  "altLabel": {
    "nb": [
      "Åbor"
   ]
  }
}
```

The best context I've come up with is this ([JSON-LD Playground](http://tinyurl.com/h3kboz8)):

``` json
{
  "@context": {
    "skos": "http://www.w3.org/2004/02/skos/core#",
    "prefLabel": {
      "@id": "skos:prefLabel",
      "@container": "@language"
    },
    "altLabel": {
      "@id": "skos:altLabel",
      "@container": "@language"
    }
  }
}
```

but this makes `altLabel` a set (or a js list) only if there's more than one item, an object otherwise. I would like to force `altLabel` to always be a set.

gkellogg:
Add grammar and algorithm changes to allow `@container` to be an array including `@set` along with another container type.

Fixes #269. Fixes #407.

### DIFF
--- a/spec/latest/json-ld-api/index.html
+++ b/spec/latest/json-ld-api/index.html
@@ -1067,11 +1067,18 @@
               <code>@container</code> key, which must be either
               <code>@list</code>, <code>@set</code>, <code>@index</code>,
               <span class="changed"><code>@id</code>, <code>@type</code></span>
-              or <code>@language</code>. Otherwise, an
+              or <code>@language</code>,
+              <span class="changed">
+                or an <a>array</a> containing exactly any one of those keyswords,
+                or a combination of <code>@set</code> and any of <code>@index</code>,
+                <code>@id</code>, <code>@type</code>, <code>@language</code>
+                in any order
+              </span>.
+              Otherwise, an
               <a data-link-for="JsonLdErrorCode">invalid container mapping</a>
               has been detected and processing is aborted.</li>
             <li class="changed">If processingMode is <code>json-ld-1.0</code> and the container value
-              is <code>@id</code> or <code>@type</code>, an
+              is <code>@id</code> or <code>@type</code>, or is otherwise not a <a>string</a>, an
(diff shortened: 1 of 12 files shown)


azaroth42:
:+1:

spec/latest/json-ld-api/index.html
```diff
@@ -1067,11 +1067,18 @@
               <code>@container</code> key, which must be either
               <code>@list</code>, <code>@set</code>, <code>@index</code>,
               <span class="changed"><code>@id</code>, <code>@type</code></span>
-              or <code>@language</code>. Otherwise, an
+              or <code>@language</code>,
+              <span class="changed">
+                or an <a>array</a> containing exactly any one of those keyswords,
```
niklasl: s/keyswords/keywords/

PR #509, opened 2017-06-29
Add test for compact property with @type as @id and @container as @list

dankawka:
I add these tests because probably there is a bug in jsonld.js implementation which prevents compacting fields mapped as container list with type of id.

Not working example on JsonLD playground: [link](https://json-ld.org/playground/#startTab=tab-compacted&json-ld=%7B%22%40context%22%3A%7B%22name%22%3A%7B%22%40type%22%3A%22%40id%22%2C%22%40container%22%3A%22%40list%22%2C%22%40id%22%3A%22https%3A%2F%2Fschema.org%2Fname%22%7D%7D%2C%22name%22%3A%5B%5D%7D&context=%7B%22%40context%22%3A%7B%22name%22%3A%7B%22%40type%22%3A%22%40id%22%2C%22%40container%22%3A%22%40list%22%2C%22%40id%22%3A%22https%3A%2F%2Fschema.org%2Fname%22%7D%7D%7D)

Working example on Markus Lanthaler JsonLD playground: [link](http://www.markus-lanthaler.com/jsonld/playground/?json-ld=%7B%0A%20%20%22%40context%22%3A%20%7B%0A%20%20%20%20%22name%22%3A%20%7B%0A%20%20%20%20%20%20%22%40type%22%3A%20%22%40id%22%2C%0A%20%20%20%20%20%20%22%40container%22%3A%20%22%40list%22%2C%0A%20%20%20%20%20%20%22%40id%22%3A%20%22https%3A%2F%2Fschema.org%2Fname%22%0A%20%20%20%20%7D%0A%20%20%7D%2C%0A%20%20%22name%22%3A%20%5B%5D%0A%7D&frame=%7B%7D&context=%7B%0A%20%20%22%40context%22%3A%20%7B%0A%20%20%20%20%22name%22%3A%20%7B%0A%20%20%20%20%20%20%22%40type%22%3A%20%22%40id%22%2C%0A%20%20%20%20%20%20%22%40container%22%3A%20%22%40list%22%2C%0A%20%20%20%20%20%20%22%40id%22%3A%20%22https%3A%2F%2Fschema.org%2Fname%22%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D&startTab=tab-compacted)

PR #510, opened 2017-07-01
Allow empty lists to be compacted by any term with container `@list` …

### LINKED ISSUE #508: Handling properties as @container as @list and @type as @id
Hello, while playing in playground I found a bug... probably. The problem happens when we declare field as `
"@container": "@list",
"@type": "@id"
`

You can see example here: 
[link](https://json-ld.org/playground/#startTab=tab-compacted&json-ld=%7B%22%40context%22%3A%7B%22name%22%3A%7B%22%40type%22%3A%22%40id%22%2C%22%40container%22%3A%22%40list%22%2C%22%40id%22%3A%22https%3A%2F%2Fschema.org%2Fname%22%7D%7D%2C%22name%22%3A%5B%5D%7D&context=%7B%22%40context%22%3A%7B%22name%22%3A%7B%22%40type%22%3A%22%40id%22%2C%22%40container%22%3A%22%40list%22%2C%22%40id%22%3A%22https%3A%2F%2Fschema.org%2Fname%22%7D%7D%7D)

When you provide empty array the compaction fails and we get full URI instead of `abc`. If we provide at least 1 value, the compaction works fine and we get `abc` property as array with provided values which is expected.

Expected result after compaction empty array would be empty array, rather than expanded form.

gkellogg:
…that matches the property IRI:

* Update inverse context to creation to include an `@any` member for the _container_ member of _container map_ for a given _iri_.
* Update IRI Compaction to set _type/language_ to `@any` if value is an empty list object.

Fixes #508. PR #509 is appropriate with this change.

### DIFF
--- a/spec/latest/json-ld-api/index.html
+++ b/spec/latest/json-ld-api/index.html
@@ -2409,10 +2409,13 @@
               <em>result</em> using the variable <em>container map</em>.</li>
             <li>If <em>container map</em> has no <em>container</em> member,
               create one and set its value to a new
-              <a class="changed">dictionary</a> with two members. The first member is
-              <code>@language</code> and its value is a new empty
+              <a class="changed">dictionary</a> with <span class="changed">three</span> members.
+              The first member is <code>@language</code> and its value is a new empty
               <a class="changed">dictionary</a>, the second member is <code>@type</code>
-              and its value is a new empty <a class="changed">dictionary</a>.</li>
+              and its value is a new empty <a class="changed">dictionary</a>,
+              <span class="changed">and the third member is <code>@any</code>
+                and its value is a new <a>dictionary</a> with the member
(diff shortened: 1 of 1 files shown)


gkellogg:
We might consider this as an erratum to 1.0.

adambro:
> We might consider this as an erratum to 1.0.

@gkellogg is it actually a bug in json-ld 1.0 version? I'm asking because it's still present in digitalbazaar/jsonld.js#189 but apparently @davidlehn decided to skip tests marked for json-ld 1.1 milestone.

gkellogg:
What I meant by that is that we might consider that implementations update this behavior in their 1.0 processors outside of any plans to commit to 1.1 upgrades. That is entirely up to the implementors, and if there was a consensus agreement, we could propose to file it as a 1.0 erratum. So far, silence has been deafening; lobby your tool providers if you think this is important.

Issue #469, opened 2017-03-09
context clobbers geo: URL scheme


VladimirAlexiev:
The same issue as https://github.com/lanthaler/vCardOntology/issues/1 applies here.
When you define "geo" as a term in a context, you clobber the absolute `geo:` URI scheme.
I haven't found all places where this happens, but here are some in this repo:
- contexts/place: ` "geo": "http://schema.org/geo"`
- various examples incl http://json-ld.org/spec/latest/json-ld/#h-advanced-context-usage eg28

gkellogg:
The ability to re-define terms is a feature, not a bug. In this regard, it's not unlike the scoping of variables in modern programming languages.

Please provide the corrective action you believe is necessary, or close the issue if you're satisfied with this response.

gkellogg:
Further more on your issue, this is not unlike defining an `http` or `https` term, which can have the same behavior. You can always undefine a term in a more local context as follows:

```
{
  "@context": {
    "geo": null
  }
}
```

VladimirAlexiev:
Let me ask this: would you be happy if W3C published ANY context officially that included terms `http`, `file` or `mailto`? How's the `geo` URI scheme inferior to those three?

>scoping of variables in modern programming language

Yes, JSONLD terms and Turtle prefixes are similar to local var names. But they are more, since you cannot practically work with full URLs without giving them names (prefixes).
- Standardizing those names is a GOOD THING, eg @cygri's http://prefix.cc does that for RDF prefixes. 
- Publishing official JSONLD contexts is a GOOD THING, and is almost necessary for making interoperable JS code that works with those JSONLDs.

The problem is that @lanthaler's `vcard` and @danbri's `schema` contexts have gone too far: instead of writing "vcard:geo" and "schema:geo", they facilitate the JS programmers to write just "geo". But that is ambiguous:
- it clashes with the `geo:` URI scheme 
- it clashes with the `vcard` and `schema` understanding of that term

http://prefix.cc doesn't make such mistake because it introduces short names only for whole ontologies (or classes of resources), but not for specific terms within.
- eg `dbr:` is `http://dbpedia.org/resource/`
- eg `dbc:` is `http://dbpedia.org/resource/Category:`
- but it would not accept eg `sofia:` for `http://dbpedia.org/resource/Sofia` (and it would be silly to ask for such name to be added)

[...]

PR #511, opened 2017-07-03
compactToRelative option and updates to compact IRI creation

### LINKED ISSUE #468: Add flag to disable compaction of relative URLs
When running any operation that includes compaction on a URL input, `@base` will be set to the URL. This will cause the output to convert any absolute URLs that are relative to `@base` to relative URLs, which can be undesirable in some applications.

Simply setting `base: null` in the options for these operations is an insufficient fix as these operations involve both an expansion phase and a compaction phase -- and this `null` `@base` will be applied during both phases when it is only desirable to apply it to the compaction phase. The only current workaround is to run expansion first, then run the desired operation on the expanded output (which will, unfortunately, needlessly run expansion again). Note: Perhaps the jsonld.js `skipExpansion` option can help mitigate this for simple compaction, but not for other operations.

Therefore, a flag is needed that prevents compaction to relative URLs for all operations that may include compaction (e.g. compaction, framing, flattening).

Historical discussion on this feature: https://github.com/json-ld/json-ld.org/issues/204
Related jsonld.js issue: https://github.com/digitalbazaar/jsonld.js/issues/171

gkellogg:
* Added `compactToRelative` option and tests. Fixed test vocabulary to match reality. Fixes #468.
* Change algorithm for creating compact IRIs to favor terms ending with colon `':'` (in which case no colon is added) or end with a gen-delim and require that such terms be simple terms.

### DIFF
--- a/spec/latest/json-ld/index.html
+++ b/spec/latest/json-ld/index.html
@@ -1043,12 +1043,11 @@
     underscore (<code>_</code>), the value is interpreted as <a>blank node identifier</a>
     instead.</p>
 
-  <p class="changed">In JSON-LD 1.1, terms will be used as compact IRI prefixes
+  <p>In JSON-LD 1.1, terms will be used as <a>compact IRI</a> prefixes
     when compacting only if
-    they are <a>simple term definitions</a> and either end with a URI <a data-cite="RFC3986#section-2.2">gen-delim</a> character (e.g,
-    <code>/</code>, <code>#</code> and others, see RFC3986) or where the term
-    itself ends with a colon (<code>:</code>). When compacting, preference
-    is given to terms ending in a colon.</p>
+    a <a>simple term definition</a> is used where the value ends with a URI <a data-cite="RFC3986#section-2.2">gen-delim</a> character
+    (e.g, <code>/</code>, <code>#</code> and others, see [[!RFC3986]]),
+    or where the <a>term</a> itself ends with a colon (<code>:</code>)</p>
 
   <p>It's also possible to use compact IRIs within the context as shown in the
     following example:</p>
@@ -3943,11 +3942,11 @@ specified. The full <a>IRI</a> for
       keyword along with <code>@set</code> (other than <code>@list</code>).
       This allows a way to ensure that such property values will always
       be expressed in <a>array</a> form.</li>
-    <li>In JSON-LD 1.1, terms will be used as compact IRI prefixes
+    <li>In JSON-LD 1.1, terms will be used as <a>compact IRI</a> prefixes
       when compacting only if
-      they are <a>simple term definitions</a> and either end with a URI <a data-cite="RFC3986#section-2.2">gen-delim</a> character,
-      or where the term itself ends with a colon (<code>:</code>).
-      When compacting, preference is given to terms ending in a colon.</li>
+      a <a>simple term definition</a> is used where the value ends with a URI <a data-cite="RFC3986#section-2.2">gen-delim</a> character,
+      or where the <a>term</a> itself ends with a colon (<code>:</code>).
+      When compacting, preference is given to <a>terms</a> ending in a colon.</li>
   </ul>
 </section>
 
--- a/spec/latest/json-ld/index.html
+++ b/spec/latest/json-ld/index.html
@@ -3944,10 +3944,7 @@ specified. The full <a>IRI</a> for
   <h4>Open Issues</h4>
   <p>The following is a list of open issues being worked on for the next release.</p>
   <p class="issue" data-number="195"></p>
-  <p class="issue" data-number="269"></p>
   <p class="issue" data-number="333"></p>
-  <p class="issue" data-number="407"></p>
-  <p class="issue" data-number="429"></p>
   <p class="issue" data-number="481"></p>
 </section>
 
--- a/spec/latest/json-ld/index.html
+++ b/spec/latest/json-ld/index.html
@@ -1043,6 +1043,12 @@
     underscore (<code>_</code>), the value is interpreted as <a>blank node identifier</a>
     instead.</p>
 
+  <p class="changed">In JSON-LD 1.1, terms will be used as compact IRI prefixes
+    when compacting only if
+    they are <a>simple term definitions</a> and either end with a URI <a data-cite="RFC3986#section-2.2">gen-delim</a> character (e.g,
+    <code>/</code>, <code>#</code> and others, see RFC3986) or where the term
+    itself ends with a colon (<code>:</code>). When compacting, preference
+    is given to terms ending in a colon.</p>
 
   <p>It's also possible to use compact IRIs within the context as shown in the
     following example:</p>
@@ -2228,7 +2234,7 @@ specified. The full <a>IRI</a> for
   <p>Scoping on <code>@type</code> is useful when common properties are used to relate things of different types, where the vocabularies in use within different entities calls for different context scoping. For example, `hasPart`/`partOf` may be common terms used in a document, but mean different things depending on the context.</p>
 
   <p class="note">Scoped Contexts are a new feature in JSON-LD 1.1, requiring
-    <a>processing mode</a> set to <code>json-ld-1.1</code></p>
+    <a>processing mode</a> set to <code>json-ld-1.1</code>.</p>
 </section>
 
 
@@ -2788,7 +2794,7 @@ specified. The full <a>IRI</a> for
   </pre>
 
   <p class="note"><a>Id maps</a> are a new feature in JSON-LD 1.1, requiring
-    <a>processing mode</a> set to <code>json-ld-1.1</code></p>
+    <a>processing mode</a> set to <code>json-ld-1.1</code>.</p>
 </section>
 
 <section class="informative changed">
@@ -2869,7 +2875,7 @@ specified. The full <a>IRI</a> for
   </pre>
 
   <p class="note"><a>Type maps</a> are a new feature in JSON-LD 1.1, requiring
-    <a>processing mode</a> set to <code>json-ld-1.1</code></p>
+    <a>processing mode</a> set to <code>json-ld-1.1</code>.</p>
 </section>
 
 <section class="informative changed">
@@ -2951,7 +2957,7 @@ specified. The full <a>IRI</a> for
   </pre>
 
   <p class="note"><a>Nested properties</a> are a new feature in JSON-LD 1.1, requiring
-    <a>processing mode</a> set to <code>json-ld-1.1</code></p>
+    <a>processing mode</a> set to <code>json-ld-1.1</code>.</p>
 </section>
 
 <section class="informative">
@@ -3937,6 +3943,11 @@ specified. The full <a>IRI</a> for
       keyword along with <code>@set</code> (other than <code>@list</code>).
       This allows a way to ensure that such property values will always
       be expressed in <a>array</a> form.</li>
+    <li>In JSON-LD 1.1, terms will be used as compact IRI prefixes
+      when compacting only if
+      they are <a>simple term definitions</a> and either end with a URI <a data-cite="RFC3986#section-2.2">gen-delim</a> character,
+      or where the term itself ends with a colon (<code>:</code>).
+      When compacting, preference is given to terms ending in a colon.</li>
   </ul>
 </section>
 
--- a/spec/latest/json-ld/index.html
+++ b/spec/latest/json-ld/index.html
@@ -404,6 +404,9 @@
         This keyword is described in <a class="sectionRef" href="#named-graphs"></a>.</dd>
       <dt class="changed"><code>@nest</code></dt><dd class="changed">Collects a set of <a>nested properties</a> within
         a <a>node object</a>.</dd>
+      <dt class="changed"><code>@prefix</code></dt><dd class="changed">
+        With the value <a>true</a>, allows this <a>term</a> to be used to construct a <a>compact IRI</a>
+        when compacting.</dd>
       <dt class="changed"><code>@version</code></dt><dd class="changed">
         Used in a <a>context definition</a> to set the <a>processing mode</a>.
         New features since <a data-cite="JSON-LD-20140116">JSON-LD 1.0</a> [[!JSON-LD-20140116]] described in this specification are
@@ -1043,11 +1046,20 @@
     underscore (<code>_</code>), the value is interpreted as <a>blank node identifier</a>
     instead.</p>
 
-  <p>In JSON-LD 1.1, terms will be used as <a>compact IRI</a> prefixes
-    when compacting only if
-    a <a>simple term definition</a> is used where the value ends with a URI <a data-cite="RFC3986#section-2.2">gen-delim</a> character
-    (e.g, <code>/</code>, <code>#</code> and others, see [[!RFC3986]]),
-    or where the <a>term</a> itself ends with a colon (<code>:</code>)</p>
+  <p class="changed">In JSON-LD 1.0, terms will be used as <a>compact IRI</a> prefixes when
+  compacting only if they map to a value that ends with a URI <a
+  data-cite="RFC3986#section-2.2">gen-delim</a> character (e.g, <code>/</code>,
+  <code>#</code> and others, see [[!RFC3986]]).</p>
+
+  <p class="note">This represents a small change to the 1.0 algorithm to prevent IRIs
+    that are not really intended to be used as prefixes from being used for creating
+    <a>compact IRIs</a>.</p>
+
+  <p class="changed">When <a>processing mode</a> is set to <code>json-ld-1.1</code>, terms will be used as <a>compact IRI</a> prefixes
+    when compacting only if their <a>expanded term definition</a> contains
+    a <code>@prefix</code> member with the value <a>true</a>, or if it has a
+    a <a>simple term definition</a>  where the value ends with a URI <a data-cite="RFC3986#section-2.2">gen-delim</a> character
+    (e.g, <code>/</code>, <code>#</code> and others, see [[!RFC3986]]).</p>
 
   <p>It's also possible to use compact IRIs within the context as shown in the
     following example:</p>
@@ -3945,8 +3957,10 @@ specified. The full <a>IRI</a> for
     <li>In JSON-LD 1.1, terms will be used as <a>compact IRI</a> prefixes
       when compacting only if
       a <a>simple term definition</a> is used where the value ends with a URI <a data-cite="RFC3986#section-2.2">gen-delim</a> character,
-      or where the <a>term</a> itself ends with a colon (<code>:</code>).
-      When compacting, preference is given to <a>terms</a> ending in a colon.</li>
+      or if their <a>expanded term definition</a> contains
+      a <code>@prefix</code> member with the value <a>true</a>. The 1.0 algorithm has
+      been updated to only consider terms that map to a value that ends with a URI
+      <a data-cite="RFC3986#section-2.2">gen-delim</a> character.</li>
   </ul>
 </section>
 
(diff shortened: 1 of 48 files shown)


test-suite/tests/compact-p002-out.jsonld
```diff
@@ -0,0 +1,8 @@
+{
+  "@context": {
+    "ex": {"@id": "http://example.org/"}
+  },
+  "@id": "http://example.org/id1",
+  "@type": ["http://example.org/Type1", "http://example.org/Type2"],
+  "http://example.org/term": {"@id": "http://example.org/id2"}
+}
```
niklasl: Shouldn't `ex` be used as a prefix definition here and thus be used in the compaction? Its value ends with a "/". I would expect a (negative) test for this to leave terms as-is (as IRIs) when a simple term does *not* end with a gen-delim or the key with a ":". Or am I missing something?

spec/latest/json-ld/index.html
```diff
@@ -1043,6 +1043,12 @@
     underscore (<code>_</code>), the value is interpreted as <a>blank node identifier</a>
     instead.</p>
 
+  <p class="changed">In JSON-LD 1.1, terms will be used as compact IRI prefixes
+    when compacting only if
+    they are <a>simple term definitions</a> and either end with a URI <a data-cite="RFC3986#section-2.2">gen-delim</a> character (e.g,
```
dlongley: The term doesn't end with a URI `gen-delim` the value (string) it maps to does... so a slight rewording here and the few other places in the spec this was copied to:

`terms will be used as compact IRI prefixes when compacting only if a <a>simple term definition</a> is used and its value ends with a URI ...`

dlongley:
I really like the idea of being very explicit in one's intent to define prefixes, e.g. put a colon (`:`) at the end of your term if you want it to be used as a prefix.

However, I wonder if having two ways to specify that something should be a prefix is going too far? Why are we considering using the additional `gen-delim` in the value as a mechanism to indicate that something can be used as a prefix as well? I would suggest we just do things one way, which is that if you put a colon at the end of your term, it will be considered usable as a prefix. If you don't, it won't.

dlongley:
Regarding the `compactToRelative` flag:

First, is it set to `true` or `false` by default? My apologies if I missed that in the text, but I couldn't find it. It seemed like the algorithm would treat it as `false` by default, but the definition for the option seemed to imply it was `true` by default. I think it should be `false` because I suspect it is usually undesirable behavior -- but that's just what I've found in most (every one so far, actually) applications I've worked with.

In any case, if people disagree and think that it should be `true`, then I recommend an additional change. When calling `compact` on the API, if a URL is given for the input document, that URL is used as a default `base` for expanding the remote document prior to compaction. This makes sense. However, that same default `base` is "passed on" to the compaction phase which is quite unexpected, in my view. This behavior is documented here:

https://github.com/json-ld/json-ld.org/issues/468#issuecomment-281824217

I would like us to make a change so that the default `base` value is not passed on to the compaction phase. If compaction to a particular base is desired during compaction, this should be specified either through the `base` API option or through the use of `@base` in the target compaction context. Intent here would be clear -- and the behavior would not be unexpected.

test-suite/tests/compact-p002-out.jsonld
dlongley: As the text above is written I believe you are correct. But I've actually argued below that we shouldn't do this. I think we should specify one way to create prefixes (append a colon (`:`) to your term). I now see that you argued for this over on #469.

I also see @gkellogg's counter argument:

https://github.com/json-ld/json-ld.org/issues/469#issuecomment-285713327

However, don't these contexts need to change to call for JSON-LD 1.1 processing rules to be applied anyway? If so, why not make the simpler change?

gkellogg: @niklasl `ex` is not used, because it's not a simple term, this test is there to be sure that a term with an expanded term definition is not used for creating a compact IRI. Certainly, when expanding, it would be used, but most expanded term definitions are defined for property values, not things intended as prefixes, this the problem with defining a term "resource", and finding that when compacting, you get "resource:s", which the limitation is attempting to address.

I originally did use 1.1 processing rules for this, and we could certainly put that back in, but my worry was that 1.0 processors will continue to do things differently, and this would be confusing in any case. Also, `ex:` can certainly be used in 1.0 to create `ex::term`, which is allowed, but odd. This just fixes that behavior.

The theory is that compact IRI generation in 1.0 is basically broken, so we should attempt to fix it without invoking the heavy-handed step of needing to explicitly define this as a 1.1 processing mode.

spec/latest/json-ld/index.html
gkellogg: Yes, thanks; easy to confuse terms with their definitions.

gkellogg:
@dlongley said:

>However, I wonder if having two ways to specify that something should be a prefix is going too far? Why are we considering using the additional gen-delim in the value as a mechanism to indicate that something can be used as a prefix as well?

This try to keep compatibility with 1.0 for the most part. Adding the gen-delim restriction for an otherwise 1.0 parser, along with the limitation to use only simple term definitions, just generally makes the existing algorithms better.

> I would suggest we just do things one way, which is that if you put a colon at the end of your term, it will be considered usable as a prefix. If you don't, it won't.

I think this might be more confusing for people that don't take a closer look at 1.1 changes. Compact IRIs are used by RDF-heads, who are used to prefix definitions from Turtle, RDFa, RDF/XML that don't end in a colon. Using the colon is really intended to address corner cases. If we were to do anything different, I'd say keep the simple term definition/gen-delim changes and drop the special processing for terms ending in a colon, at least the special consideration; we could then have one processing block, and only added a colon between prefix and suffix if the prefix doesn't end with a colon. But, I'll go wherever we can reach consensus.

test-suite/tests/compact-p002-out.jsonld
dlongley: @gkellogg, ok, as an attempt to fix 1.0 I agree. Perhaps with 1.1 we can make it such that a colon must be present to define a prefix and only use the trailing gen-delim behavior in 1.0 for compatibility purposes?

gkellogg:
@dlongley said:

> First, is it set to true or false by default? My apologies if I missed that in the text, but I couldn't find it. It seemed like the algorithm would treat it as false by default, but the definition for the option seemed to imply it was true by default.

The default values of all flags are defined in the API section on JsonLdOptions, for example `compactArrays`; this is treated the same way.

The reason to default to `true` is to keep the default behavior consistent with 1.0.

> I would like us to make a change so that the default base value is not passed on to the compaction phase. If compaction to a particular base is desired during compaction, this should be specified either through the base API option or through the use of @base in the target compaction context. Intent here would be clear -- and the behavior would not be unexpected.

This is, in fact, what my implementation does: base is added to the parsed context only if `compactToRelative` is true. Interestingly, the algorithm in the compact API does not discuss what to do with base explicitly. Perhaps we need to add some wording here to explicitly say that the document base is used as the default for `@base` in the processed context used for compaction only if `compactToRelative` is true, and not otherwise. Is there some existing text I'm missing that causes the document base to be used for compaction already there?

test-suite/tests/compact-p002-out.jsonld
gkellogg: @dlongley see comment [below](https://github.com/json-ld/json-ld.org/pull/511#issuecomment-312926594). Seems like a radical departure and totally breaks otherwise fine 1.0 contexts used for creating compact IRIs which may be more confusing, but if only done in a string 1.1 mode, that may be satisfactory. Let's hear some other opinions.

dlongley:
@gkellogg,

The most important improvement we can make here, IMO, is to find a way to ensure that prefixes must be explicitly defined and, whatever mechanism we use to define them, cannot be easily confused with other usage whereby prefixes unexpectedly appear in compacted form.

I don't think it's too much of a burden to tell those familiar with RDF that, in order to define a prefix in your JSON-LD 1.1 context, you must append a colon to the term. We could also look for `"@prefix": true` in an expanded term definition if we want to keep the symmetry behind being able to transform other simple term definitions into their expanded versions, i.e.:

```json
{
  "@context": {
    "foo:": "https://example.org"
  }
}
```

Is sugar for:

```json
{
  "@context": {
    "foo:": {
      "@id": "https://example.org",
      "@prefix": true
    }
  }
}
```

If, for some reason, someone needs a term that ends in a colon that is not to be treated as a prefix, then they can do:

```json
{
  "@context": {
    "foo:": {
      "@id": "https://example.org",
      "@prefix": false
    }
  }
}
```

Or simply:

```json
{
  "@context": {
    "foo:": {
      "@id": "https://example.org"
    }
  }
}
```

As we can default to not defining a prefix when using an expanded term definition.

So, in 1.0, we address the current state of broken IRI compaction with this PR, which is essentially to only define prefixes when using simple term definitions where the value ends in gen-delim (`/`) or where the term ends in a colon (`:`). There is no way to define a prefix using an expanded term definition. If this is the intent of this PR then I'm +1 for it.

For 1.1, I think there should be only one way to use a simple term definition to define a prefix and that is to append a colon to the term (`:`). Doing so will be considered the same as using an expanded term definition with `"@prefix": true`. We could also forgo the additional keyword entirely for now and say that you must use a simple term definition to define a prefix. Of course, I'm not sure there aren't use cases where you may want to use scoped contexts or other features in conjunction with a prefix definition, but maybe authors should be forced to achieve this by using more narrow prefixed terms.

gkellogg:
@dlongley I can live with limiting terms used for generating compact IRIs to those ending in a colon when we're explicitly in 1.1 processing mode, and treat the other behavior as an errata for 1.0 and make it generally expected behavior. I send out an email to the list discussing the various options discussed here.

gkellogg:
Not a big fan of adding a `@prefix` keyword to enable this right now, though.

dlongley:
@gkellogg,

> Interestingly, the algorithm in the compact API does not discuss what to do with base explicitly. Perhaps we need to add some wording here to explicitly say that the document base is used as the default for @base in the processed context used for compaction only if compactToRelative is true, and not otherwise. Is there some existing text I'm missing that causes the document base to be used for compaction already there?

My read of the current spec is that you set the default base IRI for the active context used by the compaction algorithm using the `IRI of the currently being processed document`, which would be the IRI of the document passed as `input` to the `compact` API call. This is specified in step 3.1 of the Context Processing Algorithm:

> The base IRI of the active context is set to the IRI of the currently being processed document (which might be different from the currently being processed context), if available; otherwise to null.

However, if you implement your processor using this interpretation (as I have), you end up producing what I believe is undesirable output. The main reason applications compact a document retrieved from its URL is to make it conform to whatever context the application is using, so it needn't be concerned with any contextual details from its original form. This allows the application to define clear invariants for operation -- where every document will have the same form, regardless of its origin.

Yet, the behavior that outputs all URLs relative to the original URL bleeds some of that original context, polluting the application's otherwise tidy sandbox. It burdens the application such that it must provide its own remedy to transform URLs back to their absolute form. Or, use an expensive hack as noted in my comment. Or, deal with relative vs. absolute URLs during every relevant processing step.

If the user wants URLs made relative to a particular base, the user should specify this base in the compaction options or compaction context. A user that passed a URL to the `compact` method on a processor (to take advantage of its document fetching capability) should not be forced to do more work to undo something they didn't ask for.

This seems like broken behavior to me, but I'm willing to live with leaving it that way when `compactToRelative` is `true` if people have been relying on this. If `compactToRelative` is `false`, I would think that setting a default base wouldn't matter given how this flag guards against compacting IRIs directly in the algorithm.

gkellogg:
@dlongley said:

> I would like us to make a change so that the default base value is not passed on to the compaction phase. If compaction to a particular base is desired during compaction, this should be specified either through the base API option or through the use of @base in the target compaction context. Intent here would be clear -- and the behavior would not be unexpected.

This seems to work if we simply don't set the default base in the context if `compactToRelative` is false. Note that we can't really do this in a 1.1 mode, as we don't know the processing mode until after we've already processed the context.

Given this, with some changes to compact-r002, we can remove the check for the flag in compact IRI, as base won't be set.

test-suite/tests/compact-p002-out.jsonld
niklasl: @gkellogg I see. I wonder if using "simple" as the arbiter here (as in key + string value) is too brittle. I'll move my thinking into the main thread...

azaroth42:
Per #469, I agree with @gkellogg's comment (https://github.com/json-ld/json-ld.org/issues/469#issuecomment-285713327) that requiring `:` to force a prefix is not a good solution, sorry @dlongley, even in just the 1.1 scope. 

The gen-delims change seems okay, but I do like the idea to define the key as a prefix in the value rather than as magical sugar in the key, per https://github.com/json-ld/json-ld.org/pull/511#issuecomment-312930854 

It's 1.0 backwards compatible (if you don't understand `@prefix:true` you'll do the same thing as you did previously) and lets 1.1 context designers be clearer about their intentions.

gkellogg:
I've addressed the specific issues addressed in the PR. However, we might want to tweak the language for creating compact IRIs based on comments from @dlongley @niklasl and @azaroth42.

If we update the language in Create Term Definition to take a string _value_ and turn it into `{"@id": _value, "@prefix": false}`, we could make this more explicit, and not rely on the use of `simple term definition`, but on if `@prefix` is defined as **true** in the processed term definition. As @azaroth42 notes, this would preserve 1.0 behavior, as the `@prefix` would be ignored (and also have the side-effect of requiring `@version: 1.1` to be set). Then, to allow terms to be used for compact IRIs, you'd need to explicitly use `"@prefix": true` in the term definition.

Alternatively, we could set to **true** if the string ends with a gen-delim, which allows better backwards compatibility, but still allows a context author to avoid this by not using simple term definitions.

niklasl:
@gkellogg I think this is good progress. Here is my assessment so far.

Terms can be used as prefixes if any of these checks hold:

* Processing mode is 1.0 and the value (simple or from `@id` of an expanded definition) ends with one of the gen-delims.
* Processing mode is 1.1+ and the value is simple and ends with one of the gen-delims.
* Processing mode is 1.1+ and the expanded term definition is declared as `@prefix: true`.

Thus in 1.0, any term will work as a prefix if its value ends with a gen-delim. In 1.1, there will be a bit more control over this. By default, the term must be simple, but you can control it with the `@prefix` declaration, and that allows terms not ending in a gen-delim to be used as a prefix too.

If so, this might work. It would not require *any change* in all known cases, but (AFAICS) allow context authors to avoid the clashes we've seen.

With that, I have some more questions:

1. To be very clear, do we want a rule that effectively allows all current contexts to continue to work, even if they have the `@version: 1.1` added? The delims rule would thus be provided for that, *and* continue to be the default simple mechanism for declaring prefixes? Or is it to be considered deprecated?

2. While we're considering a new `@prefix` keyword (which I wasn't too keen on either really), why not consider having it as a top-level `@prefix` dictionary alongside `@vocab` (more akin to Turtle and RDFa)? That'd make things quite clear as it'll separate prefixes from regular terms (see also question 4). 

    Granted, with the stipulated rules above, since the current simple practise works in 99% of the time, I guess having `@prefix: true` in the term definition might be more palatable (since it isn't to be used much, and merely explicitly states something commonly implicit in the simple value ending with a prefix delimiter). But if the delims rule is to be *provisional*, a separate prefix block might actually be more clear.

3. Also, to be thorough: we *could* either restrict the delims list to just the obvious delims `#`, `/` and `:`, *or* extend it with sub-delims as well (since a namespace like `http://example.org/ns?term=` *might* be used somewhere, hidden in some enterprise system or whatnot). That still leaves out some not entirely unlikely ones like `-` and `+`. The delims rule is not waterproof, just practically likely to cause little to no real problems.

4. Should explicit prefixes also *double as terms*, or always *only expand if used as prefixes* (even with an empty local part; like in e.g. Turtle and RDFa)? I don't think we've touched on that. It might be useful in order to allow a prefix to co-exist with a property of the same name (the name "schema" comes to mind, it's not hard to imagine others).

    This might be more pure, but since *perhaps* some prefixes have been used in conjunction with terms declared as `@type: @vocab` (e.g. for owl:imports somewhere), it's not entirely a safe change either. I really think this should be thoroughly considered.

gkellogg:
@niklasl said:

> Terms can be used as prefixes if any of these checks hold:

> * Processing mode is 1.0 and the value (simple or from `@id` of an expanded definition) ends with one of the gen-delims.
> * Processing mode is 1.1+ and the value is simple and ends with one of the gen-delims.
> * Processing mode is 1.1+ and the expanded term definition is declared as `@prefix: true`.

I think this strikes the right balance. Effectively, if running in 1.0 mode, an processor can use any term that expands to a URI ending with a gen-delim (implicitly adding a `@prefix: true` in _Create Term Definition_.

In 1.1, simple terms where the definition ends in a gen-delim, get `@prefix: true` in _Create Term Definition_. An expanded term definition needs to be explicit about `@prefix`, otherwise it is treated as false.

> With that, I have some more questions:

> 1. To be very clear, do we want a rule that effectively allows all current contexts to continue to work, even if they have the `@version: 1.1` added? The delims rule would thus be provided for that, *and* continue to be the default simple mechanism for declaring prefixes? Or is it to be considered deprecated?

IMO, we can add more restrictions for 1.1, but grand-fathering the simple term definitions ending in a gen-delim makes this pretty straight-forward.

> 2. While we're considering a new `@prefix` keyword (which I wasn't too keen on either really), why not consider having it as a top-level `@prefix` dictionary alongside `@vocab` (more akin to Turtle and RDFa)? That'd make things quite clear as it'll separate prefixes from regular terms (see also question 4). 

>     Granted, with the stipulated rules above, since the current simple practise works in 99% of the time, I guess having `@prefix: true` in the term definition might be more palatable (since it isn't to be used much, and merely explicitly states something commonly implicit in the simple value ending with a prefix delimiter). But if the delims rule is to be *provisional*, a separate prefix block might actually be more clear.

Agreed; creating a separate mechanism for prefixes is overly heavy-handed. We can rely on simple `"@prefix": true` in the term definition to allow this.

> 3. Also, to be thorough: we *could* either restrict the delims list to just the obvious delims `#`, `/` and `:`, *or* extend it with sub-delims as well (since a namespace like `http://example.org/ns?term=` *might* be used somewhere, hidden in some enterprise system or whatnot). That still leaves out some not entirely unlikely ones like `-` and `+`. The delims rule is not waterproof, just practically likely to cause little to no real problems.

Overly complicated, simply create an expanded term definition for an arbitrary IRI and use `@prefix: true`.

> 4. Should explicit prefixes also *double as terms*, or always *only expand if used as prefixes* (even with an empty local part; like in e.g. Turtle and RDFa)? I don't think we've touched on that. It might be useful in order to allow a prefix to co-exist with a property of the same name (the name "schema" comes to mind, it's not hard to imagine others).

>     This might be more pure, but since *perhaps* some prefixes have been used in conjunction with terms declared as `@type: @vocab` (e.g. for owl:imports somewhere), it's not entirely a safe change either. I really think this should be thoroughly considered.

Simpler to allow them to be used as terms, but perhaps not too likely; maybe as values of properties that are defined as `@type: @vocab`, as you suggest.

gkellogg:
@dlongley, @niklasl I believe this last commit resolves your issues on the use of terms in creating compact IRIs. The `compactToRelative` issue was resolved in a previous commit.

gkellogg:
Most changes are to the API document, for which a preview can be seen [here](https://rawgit.com/json-ld/json-ld.org/issue-405-468/spec/latest/json-ld-api/index.html).

dlongley:
Introducing `@prefix` would also give us a place, in the future, to potentially indicate that a prefix is only defined for use in contexts, not in the data (for example: `@prefix: '@context'`). This would be useful for multiple reasons:

1. It is convenient (and shorter) to use prefixes in contexts -- but you may not want them used in your data.

2. Dealing with prefixes during compaction can really slow down processing; I've had to write some complex hacks to try and optimize this (https://github.com/digitalbazaar/jsonld.js/commit/a544a3a5070b9ddd52e0c49185f32c25db360507).

gkellogg:
@dlongley: Note that, at present, the `@prefix` mechanism doesn't relate to document expansion, just use for compaction. In fact, you might have an expanded term definition that has `@prefix: false`, and have it still be used for expansion.

dlongley:
@gkellogg,

Ok. I'm more concerned with its treatment (and the related performance characteristics) during compaction anyway.

PR #517, opened 2017-07-10
Add tests based on those found to have been problematic in pyld.

### LINKED ISSUE #302: Add tests for issues found in python implementation
A few bugs were fixed in the Python implementation.  They were not caught by the test suite so new test cases may be needed.  The issues are:

https://github.com/digitalbazaar/pyld/issues/24
https://github.com/digitalbazaar/pyld/issues/25
https://github.com/digitalbazaar/pyld/issues/26

gkellogg:
Fixes #302.

### DIFF
--- a/test-suite/tests/compact-0075-context.jsonld
+++ b/test-suite/tests/compact-0075-context.jsonld
@@ -1,0 +1,5 @@
+{
+  "@context": {
+    "ns": "http://example.com/core#"
+  }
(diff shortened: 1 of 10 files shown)


gkellogg:
These all pass in the playground as well as my implementation, so it should be a no-brainer. Spec is a bit vague about making IRIs relative to base, though.

PR #518, opened 2017-07-10
Mark 1.1 tests with "specVersion": "1.1", which is independent of "processingMode".

### LINKED ISSUE #489: Test suite should better support 1.0 libraries.
The test suite currently mixes 1.0 and 1.1 tests.  It is difficult to use the test suite with 1.0 libs that have not yet updated to 1.1 support.  There is a "processingMode" option that can be set to json-ld-1.0 or json-ld-1.1 but in some cases this is set to 1.0 on purpose to test 1.1 features (see tp002 and tp005 in https://github.com/json-ld/json-ld.org/blob/master/test-suite/tests/error-manifest.jsonld).  This issue might be resolved with more explicit marking of 1.1 tests separate from the processing mode.

One option might be to add a "version" option for all 1.1 tests so a 1.0 lib could skip all 1.1 tests via a simple filter.  The current "processingMode" would more directly map to the API option of the same name and could perhaps default to the version unless explicitly set otherwise.

See also:
https://github.com/digitalbazaar/jsonld.js/issues/172
https://github.com/digitalbazaar/jsonld.js/pull/179
https://github.com/json-ld/json-ld.org/commit/7374dff787690ce6f7781b1079b9023acf550234#commitcomment-21885156

gkellogg:
Update test vocabulary location to be https. Add `specVersion`.

Fixes #489.

### DIFF
--- a/test-suite/context.jsonld
+++ b/test-suite/context.jsonld
@@ -7,9 +7,9 @@
     "rdfs":     "http://www.w3.org/2000/01/rdf-schema#",
     "xsd":      "http://www.w3.org/2001/XMLSchema#",
 
-    "context":    { "@id": "input", "@type": "@id" },
+    "context":    { "@type": "@id" },
     "expect":     { "@id": "mf:result", "@type": "@id" },
-    "frame":      { "@id": "input", "@type": "@id" },
+    "frame":      { "@type": "@id" },
     "input":      { "@id": "mf:action", "@type": "@id" },
     "option":     { "@type": "@id"},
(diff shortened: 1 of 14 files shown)


gkellogg:
@davidlehn Please see if this allows you to run the test suite against an unmodified processor. It mostly does for me; basically look for `(options.specVersion || "1.0") === "1.0"`.

davidlehn:
This works.  Thanks.

Would it make sense to align `processingMode` and `specVersion` both as `"json-ld-1.1"`?  Probably not that big a deal since this is just testing framework naming.

PR #524, opened 2017-07-13
public key in hex

### LINKED ISSUE #523: Support bitcoin testnet keys
https://github.com/w3c-vc/playground/issues/3 was created in the wrong repo, should be here.  Copy pasting that and other related info here:

@ChristopherA said:

> For the DID:BTCR hackathon next week, we will be using testnet keys rather than real bitcoin keys. Testnet private keys begin with m, whereas mainnet keys begin with 1. Otherwise the crypto is the same. This playground should support both.
>
> As we will be creating test verifiable claims next week (July 10th) for the hackathon, and this should be an easy fix, can someone prioritize supporting testnet keys ASAP, or let us know if it is not possible? Thanks!

More from @ChristopherA: 

> ...the value under creator, currently specified as 
> 
>     “creator": "ecdsa-koblitz-pubkey:1LGpGhGK8whX23ZNdxrgtjKrek9rP4xWER",
> 
> Is in fact not a public key, but base58-check hash of the public key.
> 
> The private key for our demo is also a base58check encoded value of the full private key: L4mEi7eEdTNNFQEWaa7JhUKAbtHdVvByGAqvpJKC53mfiqunjBjw
> 
> However, there is no standard base58-check for public keys (a prefix such that the value starts with 4 was proposed but never implemented), but there are tools you can see that the compressed public key (we don’t used uncompressed anymore) for that private in hex 020d79074ef137d4f338c2e6bef2a49c618109eccf1cd01ccc3286634789baef4b and the base58check is 1LGpGhGK8whX23ZNdxrgtjKrek9rP4xWER
> 
> Thus to correct, this means that the default example should look like:
> 
>     “creator": "ecdsa-koblitz-pubkey:020d79074ef137d4f338c2e6bef2a49c618109eccf1cd01ccc3286634789baef4b",
> 
> or possibly:
> 
>     “creator”: “ecdsa-koblitz-pubkey-hex:020d79074ef137d4f338c2e6bef2a49c618109eccf1cd01ccc3286634789baef4b”,

harlantwood:
Attempts to fix #523.

Live version at: https://peaceful-inlet-31266.herokuapp.com/playground/

Sample testnet keypair for your testing pleasure:

`cPXLG42L3AyiKukbNdzYecS3JVC71gRhbcZEKUWNYtUXWZDnCpkQ`

`myGEzNpxqbVkyDewZuCzEmnse8jExiH2gQ`

@ChristopherA is this what you had in mind?

### DIFF
--- a/playground/playground.js
+++ b/playground/playground.js
@@ -865,7 +865,7 @@
       var signed = null;
       var publicKeyFromPrivateKey;
       try {
-        publicKeyFromPrivateKey = new bitcoreMessage.Bitcore.PrivateKey(privateKey).toPublicKey().toAddress('livenet');
+        publicKeyFromPrivateKey = new bitcoreMessage.Bitcore.PrivateKey(privateKey).toPublicKey();
       } catch (e) {
         promise = Promise.reject(e)
       }


gkellogg:
Why are we doing this on the json-ld playground, vs on a Verifiable Claims playground? I know that historically the playground has had signing, but I thought that, as this was a different concern, it was moving off of json-ld.org.

cc/ @msporny

harlantwood:
I made a comment to add this to the correct playground: https://github.com/w3c-vc/playground/issues/3#issuecomment-358150646

However, this should still be merged here, IMO, as the current behavior is wrong.

gkellogg:
I'd like to remove the BTC signing from the JSON-LD playground, as it's really out of scope for this CG. There is a VC playground now, I believe, where these things can go.

harlantwood:
Sounds right to me.  Yes there is a VC playground.

ChristopherA:
Why remove Koblitz signing? It is an important demonstration that JSON-LD can be signed by other things than PEM signatures. And it is not BTC only — many blockchains, including Ethereum, sign using Koblitz.

davidlehn:
I find it a bit odd to have any signing demos there at all.  But it was a nice place to plop down some quick examples that we wanted to show off.  It just makes the current UI cluttered and signing isn't really a core part of JSON-LD itself, just something you can do with it, like many, many other things.  Perhaps future playground work will make such things an advanced feature or move them elsewhere.

As far as this particular PR is concerned, I don't know what it does, but if it makes the demo work correctly, is there any reason not to accept it?  Removal of the feature and related discussion could be another PR.

gkellogg:
Sure, I'll merge to make this consistent, and leave removing it or referencing an alternative playground where it is available, to the future.

PR #525, opened 2017-07-21
Remove base dot segments in RFC3986 tests.

davidlehn:
Some of the [RFC3986] tests don't remove dot segments of the base URI.  The few libraries I tested do do this processing.  It's hard to tell but it might be optional in the spec.  It does seem more understandable to me to always remove all dot segments from the full output path.

[RFC3986]: https://tools.ietf.org/html/rfc3986


gkellogg:
I'll need to go back and re-check, but this test is based on others made for Turtle, RDFa and other processors, originally submitted to rdf-tests by @RubenVerborgh. See issue https://github.com/w3c/rdf-tests/issues/6 for the background. Also, see the [email in rdf-comments](https://lists.w3.org/Archives/Public/public-rdf-comments/2015Aug/0000.html) from August 2015. Basically, it indicates that URI normalization should not be applied in most cases, thus the tests including the dot segments. The algorithm in [6.3 IRI Expansion](https://www.w3.org/TR/json-ld-api/#iri-expansion) is essentially the same as used in Turtle and other RDF parsers, so I believe the same reasoning should apply. We callout the fact that normalization _should not_ occur.

> Otherwise, if document relative is true, set value to the result of resolving value against the base IRI. Only the basic algorithm in section 5.2 of [RFC3986] is used; neither Syntax-Based Normalization nor Scheme-Based Normalization are performed.

davidlehn:
Well, isn't that all confusing.  Seems it would have been easiest to just declare dot segments in base URIs as undefined behavior and not bother with tests for it.

Are there URI resolving libraries that use this behavior or do all json-ld and other rdf and similar parsers just implement custom code for this?  Does anyone have suggestions for code to look at so we can update the custom resolvers in js, python, and php libs?  I'm guessing the algorithm has to initialize the output with the unprocessed base path, then start the dot segment removal process with the relative path as input.

Unfortunate those tests didn't exist from the beginning.  It's entirely possible the digital bazaar implementations have created improper output due to this.  But who uses dot segments in base URIs anyway?

gkellogg:
[My code](https://github.com/ruby-rdf/json-ld/blob/develop/lib/json/ld/context.rb#L1030-L1032) if fairly straight-forward. The key, for me, was not to join against base if it's already an absolute IRI.

The `URI.join` code is in [RDF::URI#join](https://github.com/ruby-rdf/rdf/blob/develop/lib/rdf/model/uri.rb#L415-L445).

Testing against the margins is what it takes to ensure compatibility, which is why they were added for the other formats.

davidlehn:
I updated the jsonld.js code to handle the tests as-is.  Will update python and php libs at some point.  Closing.

PR #539, opened 2017-10-05
Updates to be consistent with test suite

### LINKED ISSUE #537: Test suite still uses http in many cases where https is no expected, due to hosing change
The suite needs to be scoured for cases where expected values shout use https scheme.  See ruby-rdf/json-ld#40

gkellogg:
now located at `https://json-ld.org/test-suite`.

Fixes #537.


davidlehn:
Looks ok.  A couple of http://example.org uses changed to https://example.org which wasn't needed but no big deal. jsonld.js tests pass. pyld fails dramatically since it's https support was already broken and now it's broken on about every test!  But that's not the test suites fault.

gkellogg:
It needed to change to https because it was used in scheme-relative tests.

gkellogg:
@davidlehn If you're good with it, can you sign off on the review?

Issue #481, opened 2017-04-12
Allow the definition of named graphs without @ keywords


RubenVerborgh:
I'd like named graphs to become first-class citizens of JSON-LD 1.1, such that it becomes really straightforward to separate things in graphs. One reason is the [separation of data, metadata, and controls](https://ruben.verborgh.org/blog/2015/10/06/turtles-all-the-way-down/), but there are several others.

If I'm not mistaken, at the moment it's not possible to define something like:
```
{
    "members": [ { "@id": "x" }, { "@id": "y" } ],
    "metadata": {
       "memberCount": 12300
    },
    "controls": {
        "nextPage": "/members/?page=2"
    }
}
````
where `metadata` and `controls` in fact correspond to different graphs.

Note also the similarly of this use case with what `application/hal+json` does with `_links` and `_embedded`.

It seems that #195 discusses something related, but that one seems specific for containers.

**Comment by dlongley** [comment]
I think this is very important for the Verifiable Claims work and would like to see it happen as well.

**Comment by RubenVerborgh** [comment]
Nanopublications are another use case.

**Comment by msporny** [comment]
Agreed. This is also very important for the Linked Data signatures work:

```json
{
    "graphData": [ ... ],
    "graphProvenanceData": { ... },
    "graphSignatureData": { ... }
}
```

... which is important for the Verifiable Claims work and the Linked Data and Distributed Ledgers work. I'd really like to see a solution for this in JSON-LD 1.1 as our current approach requires the use of '@graph', which has been balked at, which ends up being a distraction, by a number of JSON developers.

dlongley:
I've pasted a link to a patch to jsonld.js to support simple `@graph` containers here:

https://github.com/json-ld/json-ld.org/issues/195#issuecomment-328984904

gkellogg:
@dlongley can you correct that reference?

gkellogg:
Issue #549 mostly addresses this.

dlongley:
@gkellogg, is a reference broken? Not sure what you mean.

gkellogg:
No, it seemed broken when checking at TPAC, it's fine.

RubenVerborgh:
@gkellogg Thanks—could you clarify “mostly”?

gkellogg:
[...]

PR #549, opened 2017-11-16
@graph container support

davidlehn:
Work-in-progress PR.
- [x] Add `"@container": "@graph"` expand/compact tests.
- [x] Add `"@container": ["@graph", "@set"]` expand/compact tests.
- [x] Add spec text.
- [x] Add support for additional `@id` or `@index` usage with `@graph`?


gkellogg:
I'll get to my review this weekend, after completing my own implementation, thanks for doing the heavy lifting!

A concern is that, while this allows `@graph` containers, it doesn't really support _named graphs_, as the graphs created are anonymous with no possibility of reference, other than as the value of the property for which they're associated. Solving naming would requiring adding something like a `@graphName` keyword in the term definition, and could still be problematic if that name is a BNode. Probably best to address this as a simple issue, as this clearly solves an important use case, without having the ability to actually name the named graphs.

davidlehn:
@gkellogg Comments on recent commits:

I think the `@graphId` part needs a spec example or two.  At first glance of spec text and tests, I'm still not sure how it works.  Keyword should also be added to the "Syntax Tokens and Keywords" section.  Unfortunate to need a camel case keyword.  It's the only one outside of framing spec?  Potential issue is that cap-eye-dee (Id) and lower-el-dee (ld) look the same in some fonts (like right here as I type).  I could see that causing confusion.

Removing the highlight bit is unfortunate.  Would be nice to find a better solution for that.

You marked off the `@index` part as done but I think I was unclear on what I meant there.  I was suggesting perhaps we want to allow combining `@id` and `@index` with `@graph` in `@container`.  And just to be complete, allow `@set` in there too.  Not sure if this would help with the `@graphId` bit too?  So you could do things like:
```json
{
  "@context": {
    "@vocab": "ex:",
    "index-container": {
      "@id": "ex:index-container",
      "@container": ["@graph", "@index"]
    },
    "id-container": {
      "@id": "ex:id-container",
      "@container": ["@graph", "@id"]
    }
  },
  "index-container": {
    "idx1": { ... }
  },
  "id-container": {
    "ima:id": { ... }
  }
}
```
Which would expand out into something like:
```json
[
  {
    "ex:index-container": [
      {
        "@graph": ...,
        "@index": "idx1"
      }
    ],
    "ex:id-container": [
      {
        "@graph": ...,
        "@id": "ima:id"
      }
    ]
  }
]
```

gkellogg:
> I think the @graphId part needs a spec example or two.

Sure

> At first glance of spec text and tests, I'm still not sure how it works.

I'll, of course, add spec text; it's pretty straight-forward, but there may be some corner-cases to explore.

>Keyword should also be added to the "Syntax Tokens and Keywords" section.

👍 

> Unfortunate to need a camel case keyword. It's the only one outside of framing spec? Potential issue is that cap-eye-dee (Id) and lower-el-dee (ld) look the same in some fonts (like right here as I type). I could see that causing confusion.

Open to suggestions, we could simply use `@name`.

> Removing the highlight bit is unfortunate. Would be nice to find a better solution for that.

I've looked, and perhaps there's something hidden in [highlight.js](https://highlightjs.org) we could take advantage of. OTOH, it basically looks the same as it did in 1.0 (minus the diff-marking, which its just CSS).

> You marked off the @index part as done but I think I was unclear on what I meant there. I was suggesting perhaps we want to allow combining @id and @index with @graph in @container.

What is done is that `@container`: `@index` can index graphs, but does not work with the `@graph` term logic. It's of dubious use, but was pretty simple. Absent use cases, maybe we should remove it and create issues to motivate other use cases.

gkellogg:
@dlongley Thinking about it some more, it's probably better to use the `"@container": ["@id", "@graph"]` instead of introducing an `@graphId` (or perhaps `@graphName` keyword, which is not as flexible, as the name of the graph is fixed in the term definition. I'll need to spend some time on that, but (after making things consistent), I'm going to back out the `@graphId` changes and work on that.

gkellogg:
@davidlehn and @dlongley Tests, syntax and algorithms are updated to reflect my implementation of basic graph containers, as well as the map versions. Please review.

gkellogg:
@dlongley We have support for `"@graph"`, `["@graph", "@set"]`, `["@graph", "@index"]`, `["@graph", "@index", "@set"]`, `["@graph", "@id"]`, `["@graph", "@id", "@set"]`.

I'm not sure how useful `["@graph", "@index", "@set"]` and `["@graph", "@id", "@set"]`, as, unless a document had multiple graphs with the same name, this would never happen. It does complicate the language a bit, so I'm inclined to remove support for that.

The only argument to leave it, is that `@set` is valid with all other combinations, other than `@list`.

gkellogg:
Fixes #481

dlongley:
@gkellogg,

> I'm not sure how useful ["@graph", "@index", "@set"] and ["@graph", "@id", "@set"], as, unless a document had multiple graphs with the same name, this would never happen. It does complicate the language a bit, so I'm inclined to remove support for that.
> 
> The only argument to leave it, is that @set is valid with all other combinations, other than @list.

I think the `["@graph", "@index", "@set"]` case doesn't require the same name for a graph... it could just could be a collection of unnamed graphs under a single index, right?

As for `["@graph", "@id", "@set"]`, if the only advantage is consistency for consistency's sake then it seems more compelling to remove support given that it complicates language and implementations with no real use cases. Of course, if supporting the `@index` combo version and *not* the `@id` combo version would make things even more complex, then it may as well stay. It would be good to get some more opinions though given that it looks like you've already put in the work to support both.

In any case, I'm +1 to merging the PR as is and happy defer any additional decisions to later/another PR. Great work, guys, I especially like that there was no need for an additional `@graphId` or similar keyword.

gkellogg:
> As for ["@graph", "@id", "@set"], if the only advantage is consistency for consistency's sake then it seems more compelling to remove support given that it complicates language and implementations with no real use cases.

Looking at my implementation, I don't think it really makes implementation more complicated; one line in compaction, and a bit more complicated logic in error detection when looking at the term definition. The language in Syntax is slightly more complicated, but probably not worth changing now.

spec/latest/json-ld/index.html
```diff
@@ -3964,6 +4531,11 @@ <h2>Changes since 1.0 Recommendation of 16 January 2014</h2>
       a <code>@prefix</code> member with the value <a>true</a>. The 1.0 algorithm has
       been updated to only consider terms that map to a value that ends with a URI
       <a data-cite="RFC3986#section-2.2">gen-delim</a> character.</li>
+    <li>Values of properties where the associated <a>term definition</a>
+      has <code>@container</code> set to <code>@graph</code> are interpreted as
+      <a>implicitly named graphs</a>, where the associated graph name is either
+      assigned from a new <a>blank node identifier</a>, or defined using
+      <code>@graphId</code>.</li>
```
davidlehn: @gkellogg Was this supposed to be removed?

gkellogg: No, this remains, and is what is used by verifiable credentials. We may need to add more about graph id and index containers, though.

What we removed was anything about graphId.

Issue #538, opened 2017-10-01
Mixing differently typed nodes for a property


backbone87:
In the spec under point 6.2 https://json-ld.org/spec/latest/json-ld/#node-objects the following is said:
>  The values associated with keys that expand to an absolute IRI must be one of the following: [...] - an array of zero or more of the possibilities above,

I am not quite sure, if this allows something like this:
```json
{
  "http://example.com/name": [
    "Michael",
    42,
    { "@value": "Mike" }
    { "@id": "http://example.com/Michael" }
  ]
}
```

Is this kind of mixing allowed? I am quite new to the Json-LD data model and this section of the spec can be understood as "there can be multiple nodes of a property, but they must be of one type".

elf-pavlik:
Unfortunately https://json-ld.org/playground/ seems broken at this moment but you can also test your snippets with http://linter.structured-data.org/ You just need to surround it with `<script type="application/ld+json"></script>`
```html
<script type="application/ld+json">
{
  "http://example.com/name": [
    "Michael",
    42,
    { "@value": "Mike" },
    { "@id": "http://example.com/Michael" }
  ]
}
</script>
```

As far as I can tell your snippet looks like valid JSON-LD. Once JSON-LD playground starts working again you can use it to see expanded version as well as N-Quads to make sure that you get intended results. Meanwhile you can also try out any of the JSON-LD compliant implementations listed on https://json-ld.org/#developers

gkellogg:
You can also try [my distiller](http://rdf.greggkellogg.net/distiller?command=expand&format=jsonld&output_format=jsonld), which is up-to-date with the 1.1 JSON-LD algorithms.

In this case, yes, a property (expanding to an absolute IRI) can have multiple values, consisting of value-objects, node-objects, or other JSON scalar types, which are interpreted based on possible context information associated with the property.

The distiller uses the same code as the linter, but accepts a wider variety of input formats.

backbone87:
Thanks for the clarification. Idk, if its worth changing the wording in the spec to make this one more obvious. For me as a non native english speaker, it was not immediately clear, how this statement was meant.
If you fell alright about the wording, feel free to close this one

gkellogg:
I'm afraid it seems pretty clear to me, as it's quite explicit about exactly what the value of such a key can be. Please suggest wording that you believe would improve it.

Alternatively, we could add an example of a property with such heterogenous values.

azaroth42:
:+1: to adding an example with differently typed values.

PR #554, opened 2017-12-07
Clarify that multiple property values may be of different types

gkellogg:
Fixes Issue 538.

### DIFF
--- a/spec/latest/json-ld/index.html
+++ b/spec/latest/json-ld/index.html
@@ -2466,7 +2466,7 @@ specified. The full <a>IRI</a> for
       creating an <a>implicitly named graph</a>. When expanded, these become
       <a>simple graph objects</a>.</p>
 
- <p>An alternative to our example above could use an anoymously named graph
+ <p>An alternative to our example above could use an anoymously <a>named graph</a>
     as follows:</p>
 
     <pre class="example nohighlight" data-transform="updateExample"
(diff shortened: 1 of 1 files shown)


gkellogg:
@backbone87 see if this helps clarify the language.

spec/latest/json-ld/index.html
```diff
@@ -2466,7 +2537,7 @@ <h3>Graph Containers</h3>
       creating an <a>implicitly named graph</a>. When expanded, these become
       <a>simple graph objects</a>.</p>
 
- <p>An alternative to our example above could use an anoymously named graph
+ <p>An alternative to our example above could use an anoymously <a>named graph</a>
```
dlongley: typo: anoymously => anonymously

backbone87:
Regarding #538 : very clear now

gkellogg:
We used "named graph" throughout the document, as that is the data model construct we're drawing upon. A more pervasive change might be to create the term "subgraph", and define it in terms of the RDF "named graph".

Similar is the use of "graph name", vs graph label, and the fact that even unlabeled subgraphs logically still have a name/label.

However, this is was cleaned up text and the issue/PR is about clarifying multiple heterogeneous values of a property, and not really about named graphs. A new issue about creating an abstraction level above "named graphs" is probably a better way to go.

gkellogg:
I went ahead and did some improvement of unnamed named graphs. Anything else is a big change, which should be discussed. Personally, I think it's use among non-expert users is limited, and an understanding of the underlying RDF data model is important in such cases.

PR #557, opened 2017-12-10
Issue 555 improve embedding examples

gkellogg:
@elf-pavlik see if this change satisfies your needs.

### DIFF
--- a/spec/latest/json-ld/index.html
+++ b/spec/latest/json-ld/index.html
@@ -1413,26 +1413,54 @@ the data type to be specified with each piece of data.</p>
     <a>property</a> values. This is a commonly used mechanism for
     creating a parent-child relationship between two <a>nodes</a>.</p>
 
-  <p>The example shows two nodes related by a property from the first node:</p>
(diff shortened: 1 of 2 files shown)


elf-pavlik:
Yes! Thank you @gkellogg much appreciated :pray:

spec/latest/json-ld/index.html
```diff
@@ -1413,26 +1413,54 @@ <h2>Embedding</h2>
...
+      "@vocab": "http://schema.org/",
+      ****"knows": {"@type": "@id"}****
+    },
+    "name": "Manu Sporny",
+    "@type": "Person",
+    ****"knows": "http://foaf.me/gkellogg#me"****
+  }, {
+    "@id": "http://foaf.me/gkellogg#me",
+    "@type": "Person",
+    "name": "Gregg Kellogg"
```
elf-pavlik: I took another look at the latest rendered spec with this update and noticed that the `@context` from other object doesn't seem to apply here and `name` doesn't have mapping to IRI. Does this object need something like `"@context": { "@vocab": "http://schema.org/" }`?

PR #559, opened 2017-12-12
Fix expansion algorithm to not reject a node containing only `@id`

### LINKED ISSUE #532: bug in frame expansion algorithm?
I have a frame that just specifies an id:

```json
{
  "@id": "http://purl.bdrc.io/resource/W30020"
}
```

as I just want this ID to be framed. It sounds quite legit to me, but when [expanding the frame](https://json-ld.org/spec/latest/json-ld-api/#expansion-algorithms), the result of the expansion is null because of `13.2`. Maybe this could be true only if `frame expansion` flag is not present?

gkellogg:
when framing.

Fix frame matching algorithm to not match a node if frame has a non-empty `@id`, which is not an empty dictionary.
Fixes #532.

### DIFF
--- a/spec/latest/json-ld-api/index.html
+++ b/spec/latest/json-ld-api/index.html
@@ -1872,7 +1872,7 @@
         an <a>active property</a>, and a <em>value</em> to expand.</p>
 
       <ol class="algorithm">
-        <li></span>If the <a>active property</a> has a <a>type mapping</a>
+        <li>If the <a>active property</a> has a <a>type mapping</a>
           in <a>active context</a> that is <code>@id</code>,
           <span class="changed">and the <em>value</em> is a <a>string</a>,</span>
           return a new
--- a/spec/latest/json-ld-api/index.html
+++ b/spec/latest/json-ld-api/index.html
@@ -1817,7 +1817,10 @@
               the keys <code>@value</code> or <code>@list</code>, set <em>result</em> to
               <code>null</code>.</li>
             <li>Otherwise, if <em>result</em> is a <a class="changed">dictionary</a> whose only
-              key is <code>@id</code>, set <em>result</em> to <code>null</code>.</li>
+              key is <code>@id</code>, set <em>result</em> to <code>null</code>.
+              <span class="changed">
+                When the <code>frame expansion</code> flag is set, a <a>dictionary</a>
+                containing only the <code>@id</code> key is retained.</span></li>
           </ol>
         </li>
         <li>Return <em>result</em>.</li>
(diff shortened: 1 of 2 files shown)


gkellogg:
cc/ @eroux

spec/latest/json-ld-api/index.html
```diff
@@ -1817,7 +1817,10 @@ <h3>Algorithm</h3>
               the keys <code>@value</code> or <code>@list</code>, set <em>result</em> to
               <code>null</code>.</li>
             <li>Otherwise, if <em>result</em> is a <a class="changed">dictionary</a> whose only
-              key is <code>@id</code>, set <em>result</em> to <code>null</code>.</li>
+              key is <code>@id</code>, set <em>result</em> to <code>null</code>.
+              <span class="changed">
+                When the <code>frame expansion</code> flag is set, a <a>dictionary</a>
+                containing only the <code>@id</code> key is retained.</span></li>
```
dlongley: Missing closing tag for `</span>`?

spec/latest/json-ld-framing/index.html
```diff
@@ -1089,7 +1089,10 @@ <h2>Frame Matching Algorithm</h2>
   of a set of properties, if neither <code>@id</code>, nor <code>@type</code> are in the frame.</p>
 
 <ol class="algorithm">
-  <li><em>Node</em> matches if it has an <code>@id</code> property including any <a>IRI</a> or <a>blank node</a> in the <code>@id</code> property in <em>frame</em>.
+  <li><em>Node</em> matches if it has an <code>@id</code> property
+    including any <a>IRI</a> or <a>blank node</a> in the <code>@id</code>
+    property in <em>frame</em>. Otherwise, if <em>frame</em> has a non-empty <code>@id</code> property
+    which is not an empty <a>dictionary</a>, <em>node</em> does not match.
```
dlongley: I had to read this a few times to get it, but I do now. It's essentially, do the first sentence to determine if the node definitely matches. Then, if it doesn't, you don't know yet if the node matches via some other rules. But you still need to check if it should be withheld entirely from matching by checking to see if the frame has an `@id` that doesn't match the node's `@id`. So there's some complexity in there that isn't explicit, but is still accurate.

I think we can leave it unless others also find it hard to parse in the future.

spec/latest/json-ld-api/index.html
gkellogg: I think it's okay here, but a structure check showed a couple of other errors.

spec/latest/json-ld-framing/index.html
gkellogg: Yes, really three options, match if, do  not match if, otherwise, fall through. Maybe changing to the following is easier to parse?

> Node matches if it has an `@id` property value which also a value of the `@id` property in frame. Otherwise, node does not match, if frame has a non-empty `@id` property, other than an empty dictionary. Otherwise, frame does not have an `@id` property, continue to the next step.

eroux:
Thanks for this pull request!

PR #569, opened 2018-01-17
Maps with no keys

gkellogg:
* Allows indexing on `@none`, which does not result in adding the index to the indexed object when expanding.
* Allows compaction to use `@language`, `@index`, `@id`, and `@type` even if there is no corresponding member.


spec/latest/json-ld-api/index.html
```diff
@@ -2786,31 +2784,56 @@ <h3>Algorithm</h3>
...
+                  append the values <code>@graph@index</code> and <code>@graph@index@set</code>
+                  to <em>containers</em>.</li>
+                <li>If the value does not contain the key <code>@id</code>,
+                  append the values <code>@graph@id</code> and <code>@graph@id@set</code>
+                  to <em>containers</em>.</li>
+                <li>Append the values <code>@index</code> and <code>@index@set</code>
+                  to <em>containers</em>.</li>
+              </ol>
+            </li>
               append <code>@graph</code>, <code>@id</code>, <code>@index</code> and <code>@set</code> to <em>containers</em>.</li>
```
davidlehn: Was this line supposed to be removed due to additions above?

spec/latest/json-ld-api/index.html
```diff
@@ -2786,31 +2784,56 @@ <h3>Algorithm</h3>
...
+                      <em>containers</em>.</li>
                   </ol>
                 </li>
                 <li>Otherwise, set <em>type/language</em> to <code>@type</code>
                   and set <em>type/language value</em> to <code>@id</code>,
-                  <span class="changed">and if <a data-link-for="JsonLdOptions">processingMode</a></span>
-                  is <code>json-ld-1.1</code>, append <code>@id</code>, <code>@type</code>,
-                  and <code>@index</code> to <em>containers</em>.</li>
+                  <span class="changed">and append <code>@id</code>, <code>@id@set</code>,
+                    <code>@type</code>, and <code>@type@set</code></span>,
```
davidlehn: Should this be ordered as `@set@type`?  (Also, does that appear anywhere else in doc?)

gkellogg: Ah, yes, lexicographically.

gkellogg:
Note that that IRI Compaction introduces a number of new things to add to _containers`, due to the various permutations, and change to allow things within an index key to be indexed using `@none`.

spec/latest/json-ld/index.html
```diff
@@ -2924,30 +2927,71 @@ <h2>Data Indexing</h2>
...
+        "words": 1204
+      }****]****
+    }
+  }
+  -->
+  </pre>
+
+  <p class="changed">The special index <code>@none</code> is used for indexing
+    data which does not have an associated index, which is useful to maintain
+    a normalized represeiontation.</p>
```
davidlehn: Spelling.

spec/latest/json-ld/index.html
```diff
@@ -3169,6 +3213,30 @@ <h2>Language Indexing</h2>
   }
   -->
   </pre>
+
+  <p class="changed">The special index <code>@none</code> is used for indexing
+    data which does not have a language, which is useful to maintain
+    a normalized represeiontation.</p>
```
davidlehn: Spelling.

spec/latest/json-ld/index.html
```diff
@@ -3234,28 +3302,71 @@ <h2>Node Identifier Indexing</h2>
...
+        "words": 1204
+      }****]****
+    }
+  }
+  -->
+  </pre>
+
+  <p class="changed">The special index <code>@none</code> is used for indexing
+    <a>node objects</a> which do not have an <code>@id</code>, which is useful to maintain
+    a normalized represeiontation. The <code>@none</code> index may also be
```
davidlehn: Spelling.

spec/latest/json-ld/index.html
```diff
@@ -3437,6 +3548,52 @@ <h3>Named Graph Indexing by Identifier</h3>
 
   <p>As with <a>id maps</a>, when used with <code>@graph</code>, a container may also
     include <code>@set</code> to ensure that key values are always contained in an array.</p>
+
+  <p>As with <a>id maps</a>, the special index <code>@none</code> is used for indexing
+    <a>named graphs</a> which do not have an <code>@id</code>, which is useful to maintain
+    a normalized represeiontation. The <code>@none</code> index may also be
```
davidlehn: Spelling.

spec/latest/json-ld/index.html
```diff
@@ -3517,6 +3674,48 @@ <h2>Node Type Indexing</h2>
   -->
   </pre>
 
+  <p class="changed">The special index <code>@none</code> is used for indexing
+    <a>node objects</a> which do not have an <code>@type</code>, which is useful to maintain
+    a normalized represeiontation. The <code>@none</code> index may also be
```
davidlehn: Spelling.

spec/latest/json-ld/index.html
```diff
@@ -3437,6 +3548,52 @@ <h3>Named Graph Indexing by Identifier</h3>
 
   <p>As with <a>id maps</a>, when used with <code>@graph</code>, a container may also
     include <code>@set</code> to ensure that key values are always contained in an array.</p>
+
+  <p>As with <a>id maps</a>, the special index <code>@none</code> is used for indexing
+    <a>named graphs</a> which do not have an <code>@id</code>, which is useful to maintain
+    a normalized represeiontation. The <code>@none</code> index may also be
+    a term which expands to <code>@none</code>, such as the term <em>none</em>
```
davidlehn: Copy & paste issue: the "none" term isn't used in this example.

davidlehn:
You *almost* touched every file with this commit.
What happens when you have multiple things that would compact to `@none`?  Do they all go under `@none` key?  Or is it an error?  Should be tests one way or another and maybe spec text about it (I may have missed it).

gkellogg:
> What happens when you have multiple things that would compact to `@none`. In general, this is just fine. Two node objects without an `@id` would be indexed using `@none`, and this would round-trip. Same for index, language, and type maps.

There may be an interesting case with unamed/indexted graphs, which we should probably explore. Two anonymous graphs would seem to merge into a single graph, and would not round-trip, where you would normally consider that each anonymous graph was, in fact, named with a missing BNode. Of course, this can be controlled in the source by giving the graph an explicit blank node identifier.

I think we should explore with some test cases, and address with a note, but I don't think we need different behavior.

spec/latest/json-ld/index.html
```diff
@@ -4250,7 +4448,7 @@ <h2>Language Maps</h2>
         or an array containing both <code>@language</code> and <code>@set</code>
       </span>. The keys of a
       <a>language map</a> MUST be <a>strings</a> representing
-      [[BCP47]] language codes and the values MUST be any of the following types:</p>
+      [[BCP47]] language codes, or the <a>keyword</a> <code>@none</code>, and the values MUST be any of the following types:</p>
```
azaroth42: ..., the keyword @none or an alias that expands to it, and ...  ?

spec/latest/json-ld-api/index.html
```diff
@@ -1620,8 +1624,8 @@ <h3>Algorithm</h3>
                           and (<code>@language</code>-lowercased
                           <em>language</em>),
                           <span class="changed">unless <em>item</em> is <code>null</code></span>.
-                          <span class="ednote">From <a href="#issue-480">Issue 480</a>
-                          the <a>language tag</a> <code>@none</code> represents no language.</span></li>
+                          <span class="changed">If <em>language</em> is <code>@none</code>,
+                            or expands to <code>@none</code>, do not set the <code>@language</code> member.</span>
```
azaroth42: Do we need explicit mention that the underlying RDF datatype is different? (xsd:string versus rdf:langString)

azaroth42:
I agree that having an array of objects indexed to `@none` is fine. An unnamed named graph that gets put into an index seems like a pathological edge case, that anyone who knows what they're doing would instinctively avoid anyway.

gkellogg:
@azaroth42 and @davidlehn your concerns have been addressed.

spec/latest/json-ld-api/index.html
azaroth42: Right, I agree that @none is json-ld flavored sugar... but the RDF Concepts spec asserts different datatypes based on the presence or lack of a language tag. At least that's my understanding of https://www.w3.org/TR/rdf11-concepts/#section-Graph-Literal

If there's a language tag, then the data type is necessarily rdf:langString. And if there isn't, then it's not. If there isn't another data type, it's an xsd:string.  

Which we can be completely silent about, but I'd like us to have thought about that silence :)

gkellogg: [Language maps](https://json-ld.org/spec/latest/json-ld/#language-maps) have a proscribed form, where the key is a string, and the value is either a string or an array of strings, so the expansion of a map value of `@none`, would be like `{"@value": "non-languaged-tagged value"}`, thus no datatype or language. The [Object to RDF Conversion](https://json-ld.org/spec/latest/json-ld-api/#object-to-rdf-conversion) algorithm renders this as a literal without datatype or language, which, as you note, is semantically equivalent to having the datatype `xsd:string`.

Do you think this needs some exposition in either document?

niklasl:
Is this feature compatible with an approach possible in 1.0, where you'd use two distinct terms to separate language-keyed values and non-language-tagged ones? Like:
```json
{
  "@context": {
   "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
   "xsd": "http://www.w3.org/2001/XMLSchema#",
   "labelByLang": {"@id": "rdfs:label", "@container": "@language"},
   "labelString": {"@id": "rdfs:label", "@language": null}
  },
  "labelByLang": {"en": "thing", "sv": "sak"},
  "labelString": "stuff"
}```

**Comment by gkellogg** [comment]
We may need more tests for this, and the IRI Compaction language may need to be updated. In most cases, preference in term selection is given to the most specific match; there are tests for this, but we may not have sufficient overlap. The language in 2.9.3 may need to be moved until after `@none` is added in 2.10 to ensure this.

I'll look through to make sure we have adequate test coverage for these various cases.

On a personal note, I have long thought the term selection logic to be over complicated in trying to maintain such variation in matching, which is not the typical case. In 99% of cases (I would guess) there is a single term matching an IRI, and that would be the one to use. Allowing for matching from among a set of possible IRIs complicates the algorithm, and is one aspect of performance issues for large datasets.

**Comment by gkellogg** [comment]
CG resolved to merge: https://json-ld.org/minutes/2018-01-22/index.html#resolution-4

**Comment by niklasl** [comment]
If further work on this is to be done (apart from ensuring the [explicit none-term case](#issuecomment-359481198) works),  could we consider this to be an opt-in feature for map containers; e.g. by declaring `"@container": ["@language", "@none"]`? (Although I understand the desire to get away from complex term selection rules, unless we're to evict some of the existing mechanics, we need to support it going forward.)

The "none-term" case is important, we've used it for the National Library data in Sweden (since, in library data in general, lacking language tags is the norm, alas.)

**Comment by gkellogg** [comment]
@niklasl Can you provide an example which illustrates your concern. I'd hate to have to add `@none` as a container value, and it's not clear how that would fit anyway, as `@none` is effectively no container.

You can inhibit the behavior by defining another term which would be used for terms which you don't want to be in a language map, as the test case added does.

As I mentioned on the call, we could also vary based on `processingMode`/`version`, and only allow `@none` in a language map if `version` is 1.1 (or processingMode is `json-ld-1.1`). However, the ability to define a term that is matched with a higher priority than language maps for values without a language would seem to adequately address this issue for me.

**Comment by niklasl** [comment]
@gkellogg Indeed, if the behaviour is inhibited to handle the example I provided above, the critical part of the issue is solved, since anyone using language-maps in 1.1 can declare a "companion" term for the none-case.

Apart from the reasonably undesired full IRI fallback, there is still one case which works in 1.0 that isn't compatible with this IIUC. If we use `@vocab`, in 1.0, this will round-trip:

```json
{
  "@context": {
   "@vocab": "http://www.w3.org/2000/01/rdf-schema#",
   "labelByLang": {"@id": "label", "@container": "@language"}
  },
  "labelByLang": {"en": "thing", "sv": "sak"},
  "label": "stuff"
}
```
Whereas with none-maps in 1.1, we now get only ` "labelByLang": {"en": "thing", "sv": "sak", "@none": "stuff"}`.

The new behaviour is reasonably a better default (and I am very grateful for your work on implementing it!). It is technically is a breaking change though (as shown in this and the above example). I think we need to document it in tests, and probably add the processingMode >= 1.1 check as you suggest. And also add clear instructions for how to upgrade to 1.1, by stating that you now have to use a companion term for language-maps in order to single out the non-language-tagged values using distinct terms.

I'm not keen on my opt-in either, I just wanted to mention it in case we end up not being able to justify the break. (And if so, while `@none` isn't a reasonable container value (in that you cannot use `@container: @none`), technically, it could be, just a pointless one (a map with one key (`@none`) where all values for the term are put...). Hopefully that is moot of course.)

gkellogg:
See #571.

PR #573, opened 2018-01-26
Update relative IRI definition

### LINKED ISSUE #488: Properties can not be relative IRIs -- requires better documentation
It recently came as a surprise to me that, in JSON-LD, you can not use a relative IRI as a property. In [this example](http://json-ld.org/playground/#startTab=tab-expanded&json-ld=%7B%22%40context%22%3A%7B%22%40base%22%3A%22http%3A%2F%2Fexample.org%2F%22%2C%22foo%22%3A%22http%3A%2F%2Fexample.org%2Ffoo%22%7D%2C%22foo%22%3A%7B%22%40id%22%3A%22.%2Fbar%22%7D%2C%22.%2Fbaz%22%3A%22hello%20world%22%7D), the key ``"./baz"`` is ignored (although ``"./bar"`` is correctly interpreted as a value of ``@id``).

Then I realized that this is probably a feature, not a bug, and I can see the rationale. After digging into the spec, I also realized that it is indeed documented in the Expansion algorithm (it all comes down to *document relative* defaulting to ``false`` in the IRI Expansion function).

However, I think other people might get confused by it, especially because nothing in the JSON-LD syntax document hints at this. On the contrary, section [5.2](https://www.w3.org/TR/json-ld/#iris) states

> In JSON-LD, IRIs may be represented as an absolute IRI or a relative IRI.

And Section [8.2](https://www.w3.org/TR/json-ld/#node-objects) states:

> All keys which are not IRIs, compact IRIs, terms valid in the active context, or one of the following keywords MUST be ignored when processed:

Ironically, in many other places of the spec, absolute IRIS and relative IRIs are both explicitly mentioned as allowed (e.g. as values of ``@id`` or ``@context``).

The most obvious solution would be to replace "IRIs" by "absolute IRIS", in the sentenced quoted above. I would even argue in favor of a note, drawing the reader's attention to the fact that relatives IRIs as keys are deliberately ignored in JSON-LD.

gkellogg:
to clarify that it describes only IRIs relative to the document base, not the **vocabulary mapping**.

Add additional explanatory text on relative IRIs to distinguish them from properties, values of `@type`, or values of properties defined to be vocabulary relative.

Fixes #488.

### DIFF
--- a/spec/latest/common/terms.html
+++ b/spec/latest/common/terms.html
@@ -252,7 +252,10 @@
     <dt><dfn data-lt="triple|triples|RDF triples">RDF triple</dfn></dt><dd>
       A <a data-cite="RDF11-CONCEPTS#dfn-rdf-triple" class="externalDFN">triple</a> as specified by [[RDF11-CONCEPTS]].</dd>
     <dt><dfn data-lt="relative IRIs">relative IRI</dfn></dt><dd>
-      A relative IRI is an <a>IRI</a> that is relative to some other <a>absolute IRI</a>.</dd>
+      A relative IRI is an <a>IRI</a> that is relative to some other <a>absolute IRI</a>,
+      typically the <a>base IRI</a> of the document. Note that
+      <a>properties</a>, values of <code>@type</code>, and values of <a>terms</a> defined to be <em>vocabulary relative</em>
+      are resolved relative to the <a>vocabulary mapping</a>, not the <a>base IRI</a>.</dd>
     <dt><dfn>set object</dfn></dt><dd>
       A <a>set object</a> is a <a>JSON object</a> that has an <code>@set</code>
       member.</dd>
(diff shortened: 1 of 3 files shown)


azaroth42:
:+1:

kidehen:
What does this mean? 
Does JSON-LD now support use of IRIs for identifying the predicate of an RDF Statement, yes or no?

gkellogg:
JSON-LD has always supported IRIs for identifying predicates. The issue is document-relative, vs vocabulary-relative IRIs, which we're going to consider further. So #488 remains open.

kidehen:
Okay, I should be much clearer about the fact that my quest is for the use of "relative HTTP URIs" to identify the predicates of RDF statements when using JSON-LD notation :)

Resolving this is very important. I would like to always provide RDF-Turtle and JSON-LD examples when producing Semantic Web of Linked Data utility demos etc..

PR #577, opened 2018-01-26
Remove ActivityStreams context from contextMap.

cwebber:
That was added before AS2 had a real home for its context.  It has one
now, so this is unnecessary.

### DIFF
--- a/playground/playground.js
+++ b/playground/playground.js
@@ -46,7 +46,6 @@
   playground.contextMap = {
     // be careful when working with redirectors as as Chrome (not firefox)
     // will drop Accept: application/ld+json after redirecting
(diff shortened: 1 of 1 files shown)


playground/playground.js
```diff
@@ -44,9 +44,6 @@
 
   // map of original to modifed contexts
   playground.contextMap = {
-    // be careful when working with redirectors as as Chrome (not firefox)
-    // will drop Accept: application/ld+json after redirecting
```
davidlehn: Should leave the comments.  They (maybe) apply to future rules that might be added.

gkellogg:
Ignore the Travis failure, the work to do this is on another branch, not sure why it’s running here.

PR #486, opened 2017-05-10
Update editors, eliminate authors.

gkellogg:

PR #587, opened 2018-02-05
Update editors/authors for framing to be consistent with API document.

gkellogg:
This also relates to #486.

As the Framing document is expected to continue for the long term, this PR makes editors and authors consistent with the API document (other than moving @msporny up in the order of authorship contributions).


azaroth42:
:+1: Merging as we already agreed on #486.

PR #592, opened 2018-02-15
Snapshot 2018 02 15

gkellogg:
Snapshots of documents. Note some updates:

* Separate out Algorithm Terms from other Terms.
* Update SODT and use `json-ld11cg` as the base for shortnames, to distinguish from 1.0 work and future 1.1 WG publications.


azaroth42:
:+1:

PR #593, opened 2018-02-16
Update the Expansion Algorithm to use an explicit _frame expansion_ flag

gkellogg:
... based on the presence of the `framing` option. This allows the flag to be turned off when encountering `@default`.


gkellogg:
@davidlehn @dlongley this represents insight based on doing the jsonld.js and pyld implementations, where value objects are treated as patterns, with their property values used as arrays, except when managing `@default` values, where the array forms are invalid.

azaroth42:
Approved on CG call, 2018-02-26.

PR #594, opened 2018-02-19
Fix compaction on scoped contexts based on @type

### LINKED ISSUE #546: Allow explicitly unsetting aliases in context
If I have two contexts, one which defines an alias of `id` for `@id`, and another which does not, then each context taken separately would produce different representations for the same `uri predicate value` triple -- one with `id` and the other with `@id`.

If a block of JSON-LD with the context that does not have the definition is embedded in a block that does have it, then the alias will continue to be used, producing JSON that would be unexpected by a JSON based processor.  (A JSON-LD based processor would be fine, of course).

Example:
```
{
  "@context": { "eg": "http://example.org/", "id": "@id"},
  "id": "http://example.com/1",
  "eg:service": {
    "@context": {"foo": "http://foobar.baz/"},
    "@id": "http://example.com/2",
    "foo:note": "I want the subject to be @id, not id, in this context"
  }
}
```

Proposal:

Allow setting a key to null in a context to unset any previous definition.

Thus setting "id": null in the embedded "foo" context would produce the expected @id (assuming that the context was associated at the top level, via scoping of @type, predicate or similar)

gkellogg:
Look at term contexts from `@type` before starting to compact object members. Adds test.

Fixes #546.

### DIFF
--- a/spec/latest/json-ld-api/index.html
+++ b/spec/latest/json-ld-api/index.html
@@ -2109,8 +2109,8 @@
             <li>Return <em>result</em>.</li>
           </ol>
         </li>
-        <li>Otherwise <em>element</em> is a <a class="changed">dictionary</a>.</li>
-        <li>If <em>element</em> has an <code>@value</code> or <code>@id</code>
+        <li>Otherwise <em>element</em> is a <a class="changed">dictionary</a>.
+          If <em>element</em> has an <code>@value</code> or <code>@id</code>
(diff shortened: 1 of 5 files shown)


gkellogg:
Ruby implementation is [here](https://github.com/ruby-rdf/json-ld/commit/10f0d539c024caf10383564199a77164bcd9018c)

PR #597, opened 2018-02-26
Describe the omitGraph flag

### LINKED ISSUE #435: Provide means to get framed output without @graph
In [email discussion](https://lists.w3.org/Archives/Public/public-linked-json/2016Oct/0035.html) I (@zimeon) suggested that it is helpful to generate simple and clean framed JSON-LD, without the `@graph` structure in cases where it is not necessary, that is easy for JSON-only code to consume. 

The `@graph` structure is added in the second last step of the framing algorithm, and the [pattern suggested in 2013](https://lists.w3.org/Archives/Public/public-linked-json/2013Aug/0001.html) of `compact( frame( flat, frame), context)` to remove it again is wasteful/ugly.

@azaroth42 supports having a means to avoid `@graph`, suggests current behavior could remain default for backwards compatibility.

@gkellogg recalls _"the thought at the time was that a frame could return one or more results, and it would be odd if the results had a different shape depending. However, in my experience, when a frame is used in this context, the user usually knows to expect just a single result"_ and suggests _"Rather than add yet another flag, I would suggest that if the outermost frame in a frame document contains `@graph`, that we ensure that the result does as well, otherwise, only use it if necessary to contain multiple results."_

gkellogg:
to determine if `@graph` is always used for framing results, or only if necessary.

Note that the default behavior depends on the processing mode.

Fixes #435


azaroth42:
Perhaps it's worth flipping the flag to make it easier to understand and make people explicitly get rid of `@graph` rather than it disappearing magically, compared to 1.0?

e.g.  

includeGraph  (default true)

If the includeGraph flag is set to false and the processing version is 1.1 or higher, then the "@graph" key will only be included if necessary.

Otherwise I worry that something like the following might occur: Someone might define a 1.0 frame and use it on a document with a 1.0 context that happens to embed a 1.1 context by reference, which resets the processing mode, and all of a sudden @graph disappears.

gkellogg:
Presuming that we do set the flag based on the processing mode, then perhaps. However, so much code is out there expecting that the results are under `@graph`, that we may want to consider having the default be `false`, and not depend on processing mode, in which case `omitGraph` may make more sense. Flags that default to `true` are odd (although we did that for `pruneBlankNodeIdentifiers`.

Before doing more, I'd like to get @dlongley's input.

dlongley:
@gkellogg,

I will try to look at this more closely as soon as I can -- but I wanted to say that...

> However, so much code is out there expecting that the results are under @graph, that we may want to consider having the default be false, and not depend on processing mode, in which case omitGraph may make more sense.

I'm probably going to agree with this.

azaroth42:
I'm fine with omitGraph default false (agree in retrospect that a default true flag is a bit weird), but I think it should be an error to assert it as true when the processing mode is 1.0

test-suite/tests/frame-g001-frame.jsonld
```diff
@@ -0,0 +1,13 @@
...
+    "ex": "http://example.org/vocab#"
+  },
+  "@type": "ex:Library",
+  "ex:contains": {
+    "@type": "ex:Book",
+    "ex:contains": {
+      "@type": "ex:Chapter"
+    }
+  }
+}
```
dlongley: Rest of our test files all have an EOL so these should all as well.

spec/latest/json-ld-framing/index.html
```diff
@@ -367,6 +367,31 @@ <h3>Framing</h3>
     -->
     </pre>
 
+    <p>If <a>processing mode</a> is <code>json-ld-1.1</code> or greater, or the <a>omit graph flag</a> is <code>true</code>,
```
dlongley: I think the default for the flag should be `false` given that so much code exists that expects it to be there.

test-suite/tests/frame-g001-frame.jsonld
gkellogg: @dlongley Looking at other tests, most of them do not end with a newline, for example frame-0001-frame.jsonld. The only ones I find are the toRdf-*-out.nq files, as it's a required part of the N-Quads canonical form; it's not for JSON.

We could change these, but consistency is probably best and we should either touch every test file, and include some check in CI to verify, or leave as is.

zimeon:
:+1:

gkellogg:
Resolved to pull on [2018-03-12 meeting](https://json-ld.org/minutes/2018-03-12/#resolution-2)

test-suite/tests/frame-g001-frame.jsonld
dlongley: I vote for consistency whatever that is.

PR #601, opened 2018-03-01
When compacting `@graph`, where a container type can't be used

gkellogg:
don't force value into an array.


davidlehn:
I'm not sure what the fix is in jsonld.js.  I tried removing the first arrayify part from here:
https://github.com/digitalbazaar/jsonld.js/commit/ef4f5ae82969a0a5122bfbac0e6c76591bb0a7ae

That will fix the framing tests and tests in https://github.com/json-ld/json-ld.org/pull/602, the the compact tests changed here will start failing.

gkellogg:
It was a one-line change for me, to not force _compacted value_ into an array in the Compaction Algorithm.

PR #602, opened 2018-03-01
Add tests for 1.1 graph array behavior.

davidlehn:
These test compact, though the fix to make these pass requires updates to some framing tests.


davidlehn:
See also: https://github.com/json-ld/json-ld.org/pull/601

PR #606, opened 2018-03-02
Clarify that term definitions can replace previous scoped contexts.

### LINKED ISSUE #596: Should scoped contexts be additive?
*Issue*: If two individual contexts define a term mapping with a scoped context, the second scoped context completely replaces the previous scoped context.  This is uninintuitive, given that multiple contexts are additive at the top level.

e.g.

```
{
  "@context": [
  {
    "@version": 1.1,
    "eg": "https://example.org/ns/",
    "foo": {
	  "@id": "eg:foo",
      "@context": [{
        "bar": "eg:bar"
      }]
    }
  },
    {
    "@version": 1.1,
    "foo": {
	  "@id": "eg:foo",
      "@context": [{
        "baz": "eg:baz"
      }]
    }
  },
    {
      "frob": "eg:frob"
    }
 ],
  
  "foo": {
    "bar": "1",
    "baz": "2",
    "frob": "3"
  }
}
```

The result is that `baz` and `frob` are defined but `bar` is not, as the definition of `foo` in the second block erases the definition of `bar`.

*Playground*: http://tinyurl.com/yaoxwmgf

/ht @workergnome

gkellogg:
Fixes #596.

### DIFF
--- a/spec/latest/json-ld/index.html
+++ b/spec/latest/json-ld/index.html
@@ -2317,8 +2317,8 @@ specified. The full <a>IRI</a> for
     value">values</a> of properties defined using that <a>term</a>. This allows
     values to use <a>term definitions</a>, <a>base IRI</a>,
     <a>vocabulary mapping</a> or <a>default language</a> which is different from the
-    <a>node object</a> they are contained in, in as if the
-    <a>context</a> were specified within the value itself.</p>
+    <a>node object</a> they are contained in, as if the
+    <a>context</a> was specified within the value itself.</p>
 
   <pre class="example nohighlight" data-transform="updateExample"
        title="Defining an @context within a term definition">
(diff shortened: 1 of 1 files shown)


spec/latest/json-ld/index.html
```diff
@@ -2317,7 +2317,7 @@ <h2>Scoped Contexts</h2>
     value">values</a> of properties defined using that <a>term</a>. This allows
     values to use <a>term definitions</a>, <a>base IRI</a>,
     <a>vocabulary mapping</a> or <a>default language</a> which is different from the
-    <a>node object</a> they are contained in, in exactly the same was as if the
+    <a>node object</a> they are contained in, in as if the
```
iherman: Just to be picky... shouldn't it be "<a>context</a> was specified"? Or do we always use 'context' in plural?

gkellogg:
Resolved to pull on [2018-03-12 meeting](https://json-ld.org/minutes/2018-03-12/#resolution-2)

PR #614, opened 2018-03-14
Use var instead of em for algorithm variables

gkellogg:
as ReSpec will shortly… provide a UI for highlighting variables in an algorithm.

Change CSS for algorithm to set it on the section or other containing element and not repeat on each ol, which aids in finding all variables in an algorithm.
Fix up some use of term references which should be variables.


gkellogg:
This is purely editorial, but I wanted to give a short period for comment before I merge.

The ReSpec PR to support this is w3c/respec#1554.

Issue #623, opened 2018-03-27
"@container": ["@set", "@graph"] not producing a set of separate graphs


dlongley:
```js
{
  "@context": {
    "@version": 1.1,
    "credential": {"@id": "cred:credential", "@container": ["@set", "@graph"]}
  },
  "credential": [{
    "@id": "foo:1",
    "foo:bar": "1"
  }, {
    "@id": "foo:2",
    "foo:bar": "2"
  }]
}
```

N-Quads output is:

```
<foo:1> <foo:bar> "1" _:b1 .
<foo:2> <foo:bar> "2" _:b1 .
_:b0 <cred:credential> _:b1 .
```

Expected N-Quads output is:

```
<foo:1> <foo:bar> "1" _:b1 .
<foo:2> <foo:bar> "2" _:b2 .
_:b0 <cred:credential> _:b1 .
_:b0 <cred:credential> _:b2 .
```

This happens whether or not `@set` is combined with the container, i.e. if just `@graph` is used for the `@container` and more than one value is included. Each value in a `@graph` container should be a separate graph. Note that if a disjoint graph is desired, the following syntax could be used (but is also temporarily broken):

```js
{
  "@context": {
    "@version": 1.1,
    "credential": {"@id": "cred:credential", "@container": ["@set", "@graph"]}
  },
  "credential": [
    [
      {"@id": "foo:1", "foo:bar": "1"},
      {"@id":"foo:3", "foo:bar": "3"}
    ],
    {
      "@id": "foo:2",
      "foo:bar": "2"
    }
  ]
}
```

N-Quads output is:

```
<foo:1> <foo:bar> "1" _:b1 .
<foo:2> <foo:bar> "2" _:b1 .
<foo:3> <foo:bar> "3" _:b1 .
_:b0 <cred:credential> _:b1 .
```

Expected N-Quads output is:

```
<foo:1> <foo:bar> "1" _:b1 .
<foo:2> <foo:bar> "2" _:b2 .
<foo:3> <foo:bar> "3" _:b1 .
_:b0 <cred:credential> _:b1 .
_:b0 <cred:credential> _:b2 .
```

dlongley:
I filed an issue in jsonld.js as well -- though I suspect there's a bug in the algorithm and test suite.

dlongley:
It looks like the test suite does not have sufficient coverage for this case. I think we've only got tests for single values.

gkellogg:
[...]

PR #624, opened 2018-03-27
Add expand test cases for `"@container": "@graph"` with > 1 value.

dlongley:
This PR is a work in progress and here for others to check the test cases I've added.

Still need to add compact tests.


dlongley:
Addresses #623.

gkellogg:
@dlongley See what you think. It could probably use some compaction tests where there are multiple nodes in the graph, and various combinations, but I'm running out of steam right now.

PR #628, opened 2018-04-01
Limit the use of `@none` in language and index maps to 1.1 only

### LINKED ISSUE #571: Limit use of @none in maps to 1.1 mode
As discussed in https://github.com/json-ld/json-ld.org/pull/569#issuecomment-359598061, to avoid backwards-compatibility issues when compacting, the use of `@none` and in term selection should be limited to 1.1 mode.

gkellogg:
this affects only the term selection algorithm in _IRI Compaction_.

Fixes #571.

### DIFF
--- a/spec/latest/json-ld-api/index.html
+++ b/spec/latest/json-ld-api/index.html
@@ -2944,11 +2944,11 @@
               be the last <a>container mapping</a> value to be checked as it
               is the most generic.</li>
             <li class="changed">
-              If value does not contain the key <code>@index</code>, append
+              If <p>processing mode</p> is <code>json-ld-1.1</code> and value does not contain the key <code>@index</code>, append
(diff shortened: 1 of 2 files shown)


spec/latest/json-ld-api/index.html
```diff
@@ -2944,11 +2944,11 @@ <h3>Algorithm</h3>
               be the last <a>container mapping</a> value to be checked as it
               is the most generic.</li>
             <li class="changed">
-              If value does not contain the key <code>@index</code>, append
+              If <p>processing mode</p> is <code>json-ld-1.1</code> and value does not contain the key <code>@index</code>, append
```
davidlehn: "p" is the wrong tag here and below.

gkellogg:
Resolved due to https://json-ld.org/minutes/2018-04-10/#resolution-1.

PR #629, opened 2018-04-02
Remove the `pruneBlankNodeIdentifiers` option for framing

### LINKED ISSUE #627: Framed `@graph` `@container` output not properly compacted
From digitalbazaar/jsonld.js#241.

When framing, anonymous named graphs are not considered _simple graphs_, and so are not properly compacted with `"@container": "@graph"` if the `pruneBlankNodeIdentifiers` option is `false`. For example:

Input
```js
{
  "@context": {
    "@version": 1.1,
    "@vocab": "https://example.com#",
    "ex": "https://example.org/",
    "claim": {
      "@id": "ex:claim",
      "@container": "@graph"
    },
    "id": "@id"
  },
  "claim": {
    "id": "ex:1",
    "test": "foo"
  }
}
```

Frame
```js
{
  "@context": {
    "@version": 1.1,
    "@vocab": "https://example.com#",
    "ex": "https://example.org/",
    "claim": {
      "@id": "ex:claim",
      "@container": "@graph"
    },
    "id": "@id"
  },
  "claim": {}
}
```

Because of flattening, this leads to the following before compaction:

Expanded/Framed output
```js
```json
[
  {
    "@id": "_:b0",
    "http://example.org/claim": [
      {
        "@id": "_:b1",
        "@graph": [
          {
            "@id": "http://example.org/1",
            "https://example.com#test": [
              {
                "@value": "foo"
              }
            ]
          }
        ]
      }
    ]
  }
]
```

With the `pruneBlankNodeIdentifiers` option set to try, the `@id` elements are removed:
```js
[
  {
    "http://example.org/claim": [
      {
        "@graph": [
          {
            "@id": "http://example.org/1",
            "https://example.com#test": [
              {
                "@value": "foo"
              }
            ]
          }
        ]
      }
    ]
  }
]
```

This allows compaction to render the following:
```js
{
  "@context": {
    "@version": 1.1,
    "@vocab": "https://example.com#",
    "claim": {
      "@id": "ex:claim",
      "@container": "@graph"
    },
    "id": "@id"
  },
  "@graph": [
    {
      "claim": {
        "id": "ex:1",
        "test": "foo"
      }
    }
  ]
}
```

which is the expected result.

The proposed solution is to not make `pruneBlankNodeIdentifiers` an option, but the normal behavior when running in 1.1 mode. In retrospect, it was never intended that these unreferenced blank node identifiers remain, it was just a byproduct of the flattening algorithm.

gkellogg:
and make it dependent on **processing mode**.

Fixes #627.

### DIFF
--- a/spec/latest/json-ld-framing/index.html
+++ b/spec/latest/json-ld-framing/index.html
@@ -372,7 +372,7 @@ familiar with the basic RDF concepts [[!RDF-CONCEPTS]].</p>
     -->
     </pre>
 
-    <p>If <a>processing mode</a> is <code>json-ld-1.1</code> or greater, or the <a>omit graph flag</a> is <code>true</code>,
+    <p>If <a>processing mode</a> is <code>json-ld-1.1</code>, or the <a>omit graph flag</a> is <code>true</code>,
       the top-level <code>@graph</code> member may be omitted.</p>
     <pre class="example nohighlight" data-transform="updateExample"
          title="Framed library objects with omitGraph set to false">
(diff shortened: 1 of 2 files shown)


gkellogg:
Resolved due to https://json-ld.org/minutes/2018-04-10/#resolution-2.

PR #631, opened 2018-04-11
Relax to allow `@version` to be set in any context

### LINKED ISSUE #581: Multi-version processing
**Issue:**
Currently, it seems that if any context uses `@version: 1.1` then all top level contexts must be 1.1, even though 1.1 is backwards compatible with 1.0.

For example, we have a LONG list of language codes mapped to URIs, like ...
```
{"@context": {
  "en": "http://vocab.getty.edu/aat/300388277"
}}
```

This would be used with @vocab to have the language codes expand to the URIs:

```
{"@context": {
  "language": {"@id": "dct:language", "@type": "@vocab"}
}}
```
We also have a regular 1.0 context that defines a bunch of mappings:

```
{"@context": "https://linked.art/ns/v1/linked-art.json"}
```

But if I want add a replacement context to instead scope the language expansion using 1.1's [scoped contexts](https://json-ld.org/spec/latest/json-ld/#scoped-contexts) feature ...

```
{"@context": [
  "https://linked.art/ns/v1/linked-art.json",
  {"@version": 1.1,
    "language": {"@id": "dct:language", "@type": "@vocab", 
        "@context": "http://linked.art/ns/v1/languages.jsonld"}
  }]
}}
```

then it raises an error, as the top referenced context is 1.0.
This is inconsistent with the scope context (languages.json), which is also 1.0, working fine within the 1.1 context.

This seems at least not great, as there would always need to be two contexts, one for 1.0 and one for 1.1, where the only difference is that the 1.1 context declares itself as such.

**Proposal:** Allow 1.1 contexts to include 1.0 contexts by reference.

**Examples:**

  * Working 1.0 context nested inside 1.1 context:  http://tinyurl.com/ydhnvlvf 
  * Failing 1.0 context in array beside 1.1 context: http://tinyurl.com/y9hws5sw

gkellogg:
it only conflicts… if `processingMode` is set to something else in the API. Also changes term selection for compact IRIs in 1.0 mode as a necessary byproduct of this approach.

Fixes #581

### DIFF
--- a/spec/latest/json-ld/index.html
+++ b/spec/latest/json-ld/index.html
@@ -1196,7 +1196,7 @@
     URI <a data-cite="RFC3986#section-2.2">gen-delim</a> character (e.g, <code>/</code>,
     <code>#</code> and others, see [[!RFC3986]]).
     <span class="note">The previous specification allows any term to be chosen as
-      a compact IRI prefix, which lead to a poor experience.</span></p>
+      a compact IRI prefix, which led to a poor experience.</span></p>
 
   <p class="changed">In JSON-LD 1.1, terms may be chosen as <a>compact IRI</a> prefixes
     when compacting only if
--- a/spec/latest/json-ld/index.html
+++ b/spec/latest/json-ld/index.html
@@ -972,7 +972,7 @@
   </pre>
 
   <p>The first <code>context</code> encountered when processing a
-    document determines the <code>processing mode</code>,
+    document which contains <code>@version</code> determines the <code>processing mode</code>,
     unless it is defined explicitly through an API option.</p>
 
   <p class="note">Setting the <a>processing mode</a> explicitly
@@ -1191,10 +1191,18 @@
   -->
   </pre>
 
-  <p class="changed">In JSON-LD 1.0, terms will be used as <a>compact IRI</a> prefixes when
-  compacting only if they map to a value that ends with a URI <a
-  data-cite="RFC3986#section-2.2">gen-delim</a> character (e.g, <code>/</code>,
-  <code>#</code> and others, see [[!RFC3986]]).</p>
+  <p class="changed">In JSON-LD 1.0, terms may be chosen as <a>compact IRI</a> prefixes when
+    compacting only if a <a>simple term definition</a> is used where the value ends with a
+    URI <a data-cite="RFC3986#section-2.2">gen-delim</a> character (e.g, <code>/</code>,
+    <code>#</code> and others, see [[!RFC3986]]).
+    <span class="note">The previous specification allows any term to be chosen as
+      a compact IRI prefix, which lead to a poor experience.</span></p>
+
+  <p class="changed">In JSON-LD 1.1, terms may be chosen as <a>compact IRI</a> prefixes
+    when compacting only if
+    a <a>simple term definition</a> is used where the value ends with a URI <a data-cite="RFC3986#section-2.2">gen-delim</a> character,
+    or if their <a>expanded term definition</a> contains
+    a <code>@prefix</code> member with the value <a>true</a>.</p>
 
   <p class="note">This represents a small change to the 1.0 algorithm to prevent IRIs
     that are not really intended to be used as prefixes from being used for creating
@@ -5070,6 +5078,7 @@ specified. The full <a>IRI</a> for
 <section class="appendix informative">
   <h2>Changes since 1.0 Recommendation of 16 January 2014</h2>
   <ul>
+    <li>A context may contain a <code>@version</code> member which is used to set the <a>processing mode</a>.</li>
     <li>An <a>expanded term definition</a> can now have an
       <code>@context</code> property, which defines a <a>context</a> used for values of
       a <a>property</a> identified with such a <a>term</a>.</li>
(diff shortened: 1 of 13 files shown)


gkellogg:
This route seems like the most straightforward to me, but as there's really no good way to communicate expectations other than through `processingMode`, and to ensure that the order of processing doesn't affect how legacy contexts are handled, compact IRI term selection is limited to use only _simple term definitions_ in 1.0 mode, and then only when ending with an appropriate delim character.

This goes against previous discussions in the CG, but there doesn't seem to be a good alternative. What could be done, at considerably greater complexity would be to note terms after the fact if 1.1 mode is selected and mark them as not being useful for compact IRIs. If people think it's important to do this, I'll change the PR. However, note that making this change caused no other 1.0 tests to fail.

spec/latest/json-ld/index.html
```diff
@@ -1191,10 +1191,18 @@ <h2>Compact IRIs</h2>
...
-  <p class="changed">In JSON-LD 1.0, terms will be used as <a>compact IRI</a> prefixes when
-  compacting only if they map to a value that ends with a URI <a
-  data-cite="RFC3986#section-2.2">gen-delim</a> character (e.g, <code>/</code>,
-  <code>#</code> and others, see [[!RFC3986]]).</p>
+  <p class="changed">In JSON-LD 1.0, terms may be chosen as <a>compact IRI</a> prefixes when
+    compacting only if a <a>simple term definition</a> is used where the value ends with a
+    URI <a data-cite="RFC3986#section-2.2">gen-delim</a> character (e.g, <code>/</code>,
+    <code>#</code> and others, see [[!RFC3986]]).
+    <span class="note">The previous specification allows any term to be chosen as
+      a compact IRI prefix, which lead to a poor experience.</span></p>
```
azaroth42: s/lead/led/

gkellogg:
It's up to a version 1.2, or whatever, to describe how to deal with that. Certainly, if a 1.1 processor sees 1.2, it must throw it's hands up. A 1.2 processor would presumbably be able to handle 1.1. But a 2.0? there may be some intentional incompatibilities introduced at that point, which could make having both 1.1 and 2.0 be a problem, but probably not.

spec/latest/json-ld/index.html
```diff
@@ -972,7 +972,7 @@ <h2>JSON-LD 1.1 Processing Mode</h2>
   </pre>
 
   <p>The first <code>context</code> encountered when processing a
-    document determines the <code>processing mode</code>,
+    document which contains <code>@version</code> determines the <code>processing mode</code>,
```
azaroth42: This works for the current case of 1.0 being undeclared explicitly, AFAICT, as only 1.1 will be encountered, and thus it will necessarily be encountered first. If 1.0 was declared explicitly, then the 1.1 features would not be available for the scenario described in the issue.

Would it be possible to instead set it to the highest version encountered as contexts are processed, so long as the major version is the same?  Then a set of 1.1, 1.0, 1.2 would resolve until 1.2 processing, not 1.1. If a new major version is encountered, then it should probably be an error state to avoid the backwards incompatible changes.

gkellogg: The only way to set version to 1.1 is through the API; the idea was to use this as a way detect the situation of using 1.1 inadvertently. By not setting it, it is determined by the version `@version` seen, and presumed to be 1.0, if nothing is ever seen.

You raise a good point about upward compatibility, and it's not unreasonable to do that, if we don't care about detecting the use of a higher version.

In Ruby, I can declare dependencies on version numbers such as `"=1.0"`, `"~>1.0"` or `">= 1.0"`, which allow exactly 1.0, anything in 1.x which is at least 1.0 or anything greater than 1.0 respectively.

The question is, it useful to set the processor up to be restricted to 1.0 only, treat everything as 1.1, or resolve on the first/greatest value of `@version`?

gkellogg: When 1.2 comes about (if ever), it will need to determine how to handle encountering `@version: 1.1`. Right now, the only valid value of `@version` is 1.1, so for this spec, it can't happen any other way, other than through the API `processingMode` option.

davidlehn:
I'm unclear why the `compact#tp001` output changed?  Why isn't the `ex:` prefix used?

I think some more tests are needed to cover all use cases (unless other tests already cover these).  The `expand#tp001` test is added that uses a [1.0, 1.1] context pattern.  There are other patterns to test such as [1.0, 1.1, 1.0], [1.1, 1.0], [1.1, 1.0, 1.1], etc.  Practically speaking this will come up when people put a 1.0 schema.org context in some position along with contexts in other modes.

gkellogg:
@davidlehn `compact#tp001` changed because a term with an expanded term definition can no longer be used as a prefix, unless it contains `"@prefix": true`.

I'll add some tests to cover the [1.0, 1.1, 1.0], [1.1, 1.0], [1.1, 1.0, 1.1]. I think we're good on [1.0, 1.1] and on [1.1, 1.0].

spec/latest/json-ld/index.html
azaroth42: Agreed, and we might need to change it again anyway in that hypothetical future version.

PR #632, opened 2018-04-12
Clarify that scoped contexts on `@type` use the term used

### LINKED ISSUE #521: Scope context by @type: mapped term or rdf uri?
In the description of [scoped contexts by @type](https://json-ld.org/spec/latest/json-ld/#scoped-contexts), it says:

> Scoping can also be performed using a term used as a value of @type:

I believe that this is the term used for for the type's URI, not the type's URI directly? Thus if there were two separate terms, both mapped to the same type URI, they could each have different scoped mappings for the same terms and predicates.

For example:

```
{
  "AnnotationCollection": {
    "@id": "as:Collection",
    "@context": {
      "label": {
        "@id": "rdfs:label",
        "@container": ["@language", "@set"]
      }
    }
  },
  "Collection": {
    "@id": "as:Collection"
  },
  "label": "rdfs:label"
}  
```

This would allow the `Collection` to have `label` as a string, and `AnnotationCollection` to have `label` as a language map requiring an array of values for each language.

Correct?  And if so, does this need clarification in the spec?

gkellogg:
rather than it's expanded IRI, to find a scoped context.

Fixes #521.

### DIFF
--- a/spec/latest/json-ld/index.html
+++ b/spec/latest/json-ld/index.html
@@ -2405,7 +2405,7 @@ specified. The full <a>IRI</a> for
   <h2>Scoped Contexts</h2>
 
   <p>An <a>expanded term definition</a> can include a <code>@context</code>
-    property, which defines a <a>context</a> for <a data-lt="JSON-LD
+    property, which defines a <a>context</a> (an <dfn>embedded context</dfn>) for <a data-lt="JSON-LD
     value">values</a> of properties defined using that <a>term</a>. This allows
     values to use <a>term definitions</a>, <a>base IRI</a>,
     <a>vocabulary mapping</a> or <a>default language</a> which is different from the
@@ -2436,7 +2436,7 @@ specified. The full <a>IRI</a> for
 
   <p>In this case, the social profile is defined using the schema.org vocabulary, but interest is imported from FOAF, and is used to define a node describing one of Manu's interests where those properties now come from the FOAF vocabulary.</p>
 
-  <p>Expanding this document, uses a combination of terms defined in the outer context, and those defined specifically for that term.</p>
+  <p>Expanding this document, uses a combination of terms defined in the outer context, and those defined specifically for that term in an <a>embedded context</a>.</p>
 
   <pre class="example nohighlight" data-transform="updateExample"
        title="Expanded document using a scoped context">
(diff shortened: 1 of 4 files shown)


gkellogg:
Rob, please consider if the added wording clarifies this for you; it could probably be word-smithed more.

spec/latest/json-ld/index.html
```diff
@@ -2479,7 +2479,19 @@ <h2>Scoped Contexts</h2>
...
+  <p>Scoping on <code>@type</code> is useful when common properties are used to
+    relate things of different types, where the vocabularies in use within
+    different entities calls for different context scoping. For example,
+    `hasPart`/`partOf` may be common terms used in a document, but mean
+    different things depending on the context.</p>
+
+  <p>When expanding, each value of <code>@type</code> is considered
+    (ordering them lexographically) where that value is also a <a>term</a> in
+    the <a>active context</a> having its own <a>context</a>. If so, that
+    <a>context</a> as applied to the <a>active context</a>. When compacting, if
```
azaroth42: s/as/is/ I believe

spec/latest/json-ld/index.html
```diff
@@ -2479,7 +2479,19 @@ <h2>Scoped Contexts</h2>
...
-  <p>Scoping on <code>@type</code> is useful when common properties are used to relate things of different types, where the vocabularies in use within different entities calls for different context scoping. For example, `hasPart`/`partOf` may be common terms used in a document, but mean different things depending on the context.</p>
+  <p>Scoping on <code>@type</code> is useful when common properties are used to
+    relate things of different types, where the vocabularies in use within
+    different entities calls for different context scoping. For example,
+    `hasPart`/`partOf` may be common terms used in a document, but mean
+    different things depending on the context.</p>
+
+  <p>When expanding, each value of <code>@type</code> is considered
+    (ordering them lexographically) where that value is also a <a>term</a> in
+    the <a>active context</a> having its own <a>context</a>. If so, that
```
azaroth42: having its own ... subsidiary? embedded? something? ... context.

azaroth42:
Looks good overall. One typo, one minor note about clarity of wording -- I think an adjective to distinguish the scoped context of the term would help understanding.

PR #640, opened 2018-04-20
Update Base IRI to note requestURL in XHR.

### LINKED ISSUE #316: When using JSON-LD in a browser it is impossible to discover the Base IRI after a redirect
When using JSON-LD in a browser (XMLHttpRequest) it is impossible to discover the [Base IRI](http://www.w3.org/TR/json-ld/#base-iri) after a redirect because the XHR object hides the final URL of the resource.

This means documents without an explicit base URI cannot be resolved into triples in a browser

gkellogg:
Fixes #316.

### DIFF
--- a/spec/latest/json-ld-api/index.html
+++ b/spec/latest/json-ld-api/index.html
@@ -4633,7 +4633,15 @@
           <a data-link-for="JsonLdErrorCode">multiple context link headers</a>.</dd>
         <dt><dfn data-dfn-for="RemoteDocument">documentUrl</dfn></dt>
         <dd>The final URL of the loaded document. This is important
-          to handle HTTP redirects properly.</dd>
+          to handle HTTP redirects properly.
+          <span class="changed">Loading a document with a redirect MUST use the redirected URL as the <a>documentUrl</a>.</span>
+          <div class="note">If the <a>JSON-LD document</a> is retrived via <em>XMLHttpRequest</em>
+            and the server issues an HTTP redirect, the document location
+            MUST reflect the redirected location rather than that originally
+            requested. The redirected location may be provided via the <a
+            data-cite="XHR#dom-xmlhttprequest-responseurl">responseURL</a>
+            attribute; see [[XHR]].</div>
+          </dd>
         <dt><dfn data-dfn-for="RemoteDocument">document</dfn></dt>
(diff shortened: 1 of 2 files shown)


gkellogg:
cc/@graingert.

gkellogg:
@graingert said;

> This should also cover fetch

Does this indicate that there's someplace else in the spec that should be updated?

spec/latest/json-ld-api/index.html
```diff
@@ -4633,7 +4633,15 @@ <h3>RemoteDocument</h3>
...
         <dt><dfn data-dfn-for="RemoteDocument">documentUrl</dfn></dt>
         <dd>The final URL of the loaded document. This is important
-          to handle HTTP redirects properly.</dd>
+          to handle HTTP redirects properly.
+          <span class="changed">Loading a document with a redirect MUST use the redirected URL as the <a>documentUrl</a>.</span>
+          <div class="note">If the <a>JSON-LD document</a> is retrived via <em>XMLHttpRequest</em>
+            and the server issues an HTTP redirect, the document location
+            MUST reflect the redirected location rather than that originally
+            requested. The redirected location may be provided via the <a
+            data-cite="XHR#dom-xmlhttprequest-responseurl">responseURL</a>
```
elf-pavlik: I think most browsers already adopted Fetch API, maybe also worth mentioning it here?
https://fetch.spec.whatwg.org/#dom-response-url

gkellogg: @elf-pavlik care to suggest some text?

elf-pavlik: I'll double check how https://fetch.spec.whatwg.org and https://xhr.spec.whatwg.org/ stand at current time and possibly will follow up with separate PR to reference Fetch.

gkellogg: I'm getting concerned with the amount of specifics, even though in notes, with other specs. Perhaps this would be better to be more vague about using facilities such as XHR and Fetch to retrieve the redirected URL, rather than discuss specific mechanisms.

gkellogg:
I think @azaroth42 is right: we shouldn't speak to these specifics in the spec, but move to the best practices document. I suggest we close without pulling this PR and update issue #316 to put this in best practices.

PR #642, opened 2018-04-20
Add Term Selection algorithm examples.

### LINKED ISSUE #495: Question/ unexpected behavior: Why does compaction sometimes expand and not remove a prefix?
Consider the following example:

```json
{
  "@context": {
    "schema": "http://schema.org/",
    "ex": "http://example.org/",
    "name": "schema:name",
    "foo": {"@id": "ex:foo", "@type": "schema:Text"},
    "bar": {"@id": "ex:bar", "@type": "schema:Person"},
  },
  "@type": "schema:Thing",
  "name": "Jane Doe",
  "foo": "Professor",
  "bar": {
    "@type": "schema:Person",
    "name": "John Doe"
  }
}
```

Compaction under the identical context results in `bar` being expanded to `ex:bar`, but `foo` remains `foo`.  Why?  To me this is unexpected: the context seems to unambiguously define `bar` as `ex:bar`.  I think this behavior has something to do with `bar` taking a node value rather than a data value, but I fail to see why that matters.  What does the user gain by having it called `ex:bar` that wasn't clear when it was just `bar`?  

This behavior is frustrating from the standpoint of a developer consuming JSON-LD with this kind of context structure, because I feel unable to predict whether the JSON files will use `bar` or `ex:bar` (and naturally I want to just use `bar`, which is why it's explicitly mapped `bar` to `"@id": "ex:bar"` in the first place).  

Here's my example on [Playground](http://tinyurl.com/lkhutdt).

Thanks kindly for any clarification into this seemingly unexpected behavior that is no doubt more a problem of my understanding than algorithmic error!

gkellogg:
Fixes #495.

### DIFF
--- a/spec/latest/json-ld-api/index.html
+++ b/spec/latest/json-ld-api/index.html
@@ -3149,7 +3149,7 @@
 }
         </pre>
 
-        <aside class="example data-ignore" title="Language map term with language value">
+        <aside class="example" data-ignore title="Language map term with language value">
           <p>Given the member <code>{"http://example/t": {"@value": "foo", "@type": "http:/example/type"}}</code>,
             The algorithm will be invoked as follows:</p>
           <dl>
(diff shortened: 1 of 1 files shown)


gkellogg:
I didn't see a way to markedly improve the algorithm Overview, but I added an Examples section to illustrate the operation of the algorithm, that may help clarify.

cc/@cboettig.

PR #643, opened 2018-04-21
Clarify BNode pruning with keyword aliases

### LINKED ISSUE #611: blank node identifier pruning when @id aliased
It seems quite common these days to have an alias for @id in a jsonld context (see http://schema.org/docs/jsonldcontext.json ). The spec on json-ld.org [specifies](https://json-ld.org/spec/latest/json-ld-framing/#dfn-pruneblanknodeidentifiers) that:

```
pruneBlankNodeIdentifiers
    Removes @id from node objects where the value is a blank node identifier used only once within the document.
```

maybe the spec could specify that if the `@id` field is aliased in the `@context`, then `pruneBlankNodeIdentifiers` should remove the alias? Or would that be implicit?

See also https://github.com/jsonld-java/jsonld-java/issues/226

gkellogg:
* Add statement to Conformance sections of API and Framing that algorithm steps describing operations on keywords also apply to aliases of those keywords to make it clear.
* Add example using alias of `@id`.

Fixes #611.

### DIFF
--- a/spec/latest/json-ld-api/index.html
+++ b/spec/latest/json-ld-api/index.html
@@ -737,6 +737,9 @@
     so long as the end result is indistinguishable from the result that would
     be obtained by the specification's algorithms.</p>
 
+  <p>In algorithm steps that describe operations on <a>keywords</a>, those steps
+    also apply to <a>keyword aliases</a>.</p>
+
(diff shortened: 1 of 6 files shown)


gkellogg:
cc/ @eroux.

eroux:
Thanks!

PR #649, opened 2018-04-28
CG-FINAL publication snapshots for 2018-05-07.

gkellogg:
Remaining issues targeted for the "JSON-LD 1.1 CG" milestone need to be marked as deferred.


gkellogg:
I need to update the [bibliography](https://github.com/json-ld/json-ld.org/blob/master/spec/latest/common/common.js) which uses `https://json-ld.org/spec/latest/` for the location of the specs, which won't be their published locations. I'll change while generating to `https://json-ld.org/spec/CG-FINAL/json-ld/20180507/` with variations for the other specs.

davidlehn:
There's a markup problem in the generated api and framing docs with whitespace in ids and hrefs.  Perhaps a bug with respec not doing proper escaping?  For instance, the JsonLdErrorCode section has hrefs and ids like:
```html
<a href="#dom-jsonlderrorcode-colliding keywords" class="idlEnumItem">"colliding keywords"</a>
...
<dt><dfn data-dfn-for="jsonlderrorcode" data-dfn-type="dfn" id="dom-jsonlderrorcode-colliding keywords" ...><code>colliding keywords</code></dfn></dt>
```

gkellogg:
@azaroth42 these should have been published last week, and need to go on the CG Wiki.

azaroth42:
Should I merge?

gkellogg:
I'm validating before merge, I'll merge.

PR #673, opened 2018-09-02
Update term for "born" and "died" to be xsd:date not xsd:dateTime

gkellogg:
this makes it consistent with example usage.

### DIFF
--- a/contexts/person
+++ b/contexts/person
@@ -19,7 +19,7 @@
       "born":
       {
          "@id": "http://schema.org/birthDate",
-         "@type": "xsd:dateTime"
+         "@type": "xsd:date"
       },
       "child":
       {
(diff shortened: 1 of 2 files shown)


niklasl:
I'd say common practise is to just use dates for these values, making that fairly suitable for example usage, as @gkellogg says.

Note that technically it's not a restriction, since its given as schema:rangeIncludes and not rdfs:range. Also, schema:Date and schema:DateTime do not have any (RDF-)formal relation to xsd:date or xsd:dateTime. So the example context is more restrictive than schema.org, and only suitable if the *context of the data* does not vary. But that goes for many JSON-LD contexts, reasonably by design. (It makes little sense to define a restrictive context if the context isn't known...)

gkellogg:
`schema:birthDate` and `schema:deathDate` have the range of `schema:Date`, which generally corresponds to `xsd:date`. Pretty much all of our examples, and much of what is parsed, particularly for Microdata, uses XSD datatypes vs schema.org, but in any case, these are just examples.

PR #687, opened 2019-02-08
Fix incorrect context reference in error-0042

rubensworks:
As far as I understand the goal of this test, I think this test may be incorrectly defined.
Currently, jsonld.js seems to accept this input, and this fails the test.
I assume this is because the `list` context entry is not used in the JSON-LD document.

### DIFF
--- a/test-suite/tests/error-0042-in.jsonld
+++ b/test-suite/tests/error-0042-in.jsonld
@@ -1,3 +1,3 @@
 {
-  "http://example/list": [{"@list": ["foo"]}, {"@list": ["bar"]}]
+  "list": [{"@list": ["foo"]}, {"@list": ["bar"]}]
 }


gkellogg:
This is now https://w3c.github.io/json-ld-api/tests/compact-manifest.jsonld#te001 and is labeled as a 1.0-only test, as compaction to lists of lists is supported in 1.1, which may be why you're seeing jsonld.js pass this test.

The input is in expanded form, thus `http://example/list` is correct; simply `list` would fail, as it is not an IRI. The associated context will compact `http://example/list` to just `list`, and fail in 1.0, but succeed in 1.1. You can see the manifest entry in the WG test suite is marked as `specVersion: json-ld-1.0`:


    {
      "@id": "#te001",
      "@type": [ "jld:NegativeEvaluationTest", "jld:CompactTest" ],
      "name": "Compaction to list of lists",
      "purpose": "Verifies that an exception is raised in Compaction when attempting to compact a list of lists",
      "option": {"specVersion": "json-ld-1.0"},
      "input": "compact/e001-in.jsonld",
      "context": "compact/e001-context.jsonld",
      "expect": "compaction to list of lists"
    }

There should be another test for 1.1, which is expected to succeed.

rubensworks:
If I'm correct, test `error-0042` currently defines a _set of lists_, which is allowed in JSON-LD 1.0 as far as I can see (as opposed to a _list of lists_):

```jsonld
{
  "http://example/list": [{"@list": ["foo"]}, {"@list": ["bar"]}]
}
```

That is because even though the context defines `"@container": "@list"`, it has no relation to the JSON-LD document:

```jsonld
{
  "@context": {
    "list": {"@id": "http://example/list", "@container": "@list"}
  }
}
```


Could this be an incorrect expansion of the document, which instead should be the following?
```jsonld
{
  "http://example/list": {"@list": [{"@list": ["foo"]}, {"@list": ["bar"]}]}
}
```

gkellogg:
Yes, you're correct. I fixed it in the WG version of the test suite. If you like, you can update this, but I'm going to work on redirecting the tests to w3c.github.io/json-ld-api/tests.

rubensworks:
Great, thanks!

In that case I don't see the need anymore to update the test here.

PR #692, opened 2019-02-24
One playground.

davidlehn:
- Redirect old /playground-dev/ to main /playground/.
- Use newer libs for main /playground/.
- Create /playground/1.0/ snapshot.
- Create /playground/dev/ for experimentation.

Todo:
- [x] Add input options.
- [ ] Save/restore options from gists.
- [ ] Save/restore options from urls.


dlongley:
+1 for a less frustrating editor if it's easy ... but that can wait :)

PR #704, opened 2019-09-08
Add benchmark files

davidlehn:
Benchmarks previously discussed in the W3C JSON-LD WG:
https://github.com/w3c/json-ld-api/pull/59

This patch adds benchmark data here for the community to work on rather than the JSON-LD WG.

The EARL and manifest additions are a bit ad-hoc.  Comments welcome.

Comparison tool coming soon.


gkellogg:
Perhaps add a “jld:BenchmarkTest” to each test entry as well.

Also, maybe we can update the site to reference them.

Not sure what the role of expected results is for benchmark tests.

PR #774, opened 2022-05-18
+ serialize nested lists

ericprud:


ericprud:
gah, wrong upstream!
sorry for the noise

Issue #778, opened 2022-05-31
RFCs


davidlehn:
I've long had an idea for starting a JSON-LD Request For Comment "RFC" system.  Similar to PEPs, JEPs, XEPs, EIPs, and so on in other projects.  The idea being to have a collection of docs that propose and work on new features, ideas, implementation conventions, and other related items.  In some cases features may move into specs in the future.  It's worked pretty well in other communities and I imagine we could grow some new ideas here too.

I don't know what name would be best... so was just thinking of being old school with RFC.

Location would perhaps be `rfcs.json-ld.org`, so use a new repo, make some meta rfcs and indexes, and get github pages to host it.

Not sure what file formats to suggest.  Probably support both respec and markdown.  Markdown because it seems a bit more lightweight when just throwing new ideas out there.

As for topics, I've got some things queued up and it's unclear where best to document them.

- **benchmarking**: Recommendations for common implementation standards for benchmarking so we can more easily ~~compete~~ compare code.
- **custom API flags**: Recommendations for how to pass implementation specific flags to API calls.
- **safe mode**: Spec for a "safe" mode to avoid common pitfalls in use cases like digital signatures. jsonld.js is getting this feature soon and it seems worthwhile to document it and collaborate with other implementations on the idea.
- **well-formed mode**: Ideas for higher performance when you know your input is good.

I can come up with more.  I assume others can too.  Some of these things seem too small to handle with their own repos and big spec docs.  But a light weight RFC doc seems right.

Thoughts?

gkellogg:
Note that RFCs are an IETF thing, but there may be a similar concept in the W3C process. As a CG, I believe we can operate any way we want, but for things to go through REC-track, we’d need to use the WG, which does have some specific process guidelines.

So far, we’ve worked on separate specs (JSON-LD-star and JSON-LD Streaming, for example) that might be put forward as their own specs. I think what you’re calling for is a process that would serve as a way to better detail changes to existing JSON-LD Recommendations. The 2021 W3C Process document does define a “candidate addition” [1] that might be the way to look at this. I would think that the CG work towards a process that would lead to Candidate additions the WG could then drive through to create updated recommendations. (There are also “canidiate corrections”, which are closer to errata). The process for revising a Recommendation is detailed in [2].

Of course, work the CG does may never advance forward to the Working Group, and does not necessarily need to to be useful to the community. Clearly, at some point we’d need more advice on how to handle this specifically. This is a suitable topic for the upcoming call.

[...]

PR #790, opened 2022-08-24
Update jsonld js and add safe mode support.

davidlehn:
- Updates jsonld.js to 8.0.0.
- Add safe mode option UI.
- Show error details if available.  Hopefully this doesn't have degenerate cases involving huge errors or recursive serialization.  A built project would use other libraries to avoid that but this is difficult for now.
- Disables signature tabs since old jsonld-signatures no longer works with newer jsonld.js.  Bundled code and default suites are not distributed.  Need to move this to a build process if that functionality is to be kept.

### DIFF
--- a/playground/index.html
+++ b/playground/index.html
@@ -506,9 +506,9 @@ yT2IMAWbY76Bmi8TeQJAfdLJGwiDNIhTVYHxvDz79ANzgRAd1kPKPddJZ/w7Gfhm
               </a>
             </li>
             <li>
-              <a id="tab-normalized" href="#pane-normalized" data-toggle="tab" name="tab-normalized">
+              <a id="tab-canonized" href="#pane-canonized" data-toggle="tab" name="tab-canonized">
                 <i class="icon-archive"></i>
-                <span>Normalized</span>
+                <span>Canonized</span>
               </a>
             </li>
             <li>
@@ -561,8 +561,8 @@ yT2IMAWbY76Bmi8TeQJAfdLJGwiDNIhTVYHxvDz79ANzgRAd1kPKPddJZ/w7Gfhm
             <div id="pane-nquads" class="tab-pane">
               <textarea id="nquads" class="codemirror-output"></textarea>
             </div>
-            <div id="pane-normalized" class="tab-pane">
-              <textarea id="normalized" class="codemirror-output"></textarea>
+            <div id="pane-canonized" class="tab-pane">
+              <textarea id="canonized" class="codemirror-output"></textarea>
             </div>
             <div id="pane-table" class="tab-pane">
               <table id="table" class="table table-condensed">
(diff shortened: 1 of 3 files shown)


dlongley (approved):
Thanks!

We should try to get the signature stuff working again with ed25519-signature-2020 as quickly as we can.

gkellogg:
It would be useful to have a reference and note to motivate the purpose of “safe mode”.

davidlehn:
> It would be useful to have a reference and note to motivate the purpose of “safe mode”.

Yes, this is happening out of the intended order due to limited time.  This was one of the initial [RFCs](https://github.com/json-ld/json-ld.org/issues/778) I was going to write up.  I imagine there will be feedback and this first implementation will be updated as needed.

davidlehn:
> We should try to get the signature stuff working again with ed25519-signature-2020 as quickly as we can.

Arguably, the signature code should be on some other playground.  It's not strictly a json-ld thing.  In any case, the whole playground should be rewritten with modern tools at some point, and that's not a "quick" project.  And deployment would no longer be simple static files.  The only "quick" solution I've considered is to just abuse npm and npm cdns with a site specific playground build packages for the whole thing or just parts like jsigs support.

gkellogg:
> > We should try to get the signature stuff working again with ed25519-signature-2020 as quickly as we can.
> 
> Arguably, the signature code should be on some other playground.

I agree, I think the JSON-LD playground should remain focused on JSON-LD. Things that make use of the technology, such as signatures, probably should go somewhere else. If they can be hosted off of GitHub pages directly, that might make it easier to spin-up a playground in other repos.

> It's not strictly a json-ld thing. In any case, the whole playground should be rewritten with modern tools at some point, and that's not a "quick" project. And deployment would no longer be simple static files. ...

The YAML-LD group has been asking for a YAML-LD playground; a complicating factor is the limitations of the current UI (specifying YAML-LD vs JSON-LD inout or output, and perhaps allowing for N-Quads/TriG input as well as output). An updated playground architecture that allowed for changes like this more easily would be useful.

dlongley:
We should add a link to https://playground.chapi.io/ in lieu of the signature tab. That will take people to another playground that is focused on signatures involving other standards work (signing / issuing JSON-LD-based VCs).

dlongley:
A "Other Related Playgrounds" section could link to the CHAPI playground and a (future) YAML-LD playground.

playground/index.html
```diff
@@ -518,12 +554,23 @@ <h4>Public Key for Verification</h4>
...
             </div>
             <div id="pane-signed-koblitz" class="tab-pane">
               <textarea id="signed-koblitz" class="codemirror-output"></textarea>
             </div>
+            -->
+            <div id="pane-signatures" class="tab-pane">
+              <div class="alert alert-info">
+                <strong>NOTE</strong>:
+                As of 2022-08-25 the former RSA and Bitcoin signature demos are
+                temporarilly unavailable. For a demo involving JSON-LD
```
TallTed, line 569:
```suggestion
                As of 2022-08-25, the former RSA and Bitcoin signature demos are
                temporarily unavailable. For a demo involving JSON-LD
```

playground/index.html
```diff
@@ -518,12 +554,23 @@ <h4>Public Key for Verification</h4>
...
               <textarea id="signed-rsa" class="codemirror-output"></textarea>
             </div>
             <div id="pane-signed-koblitz" class="tab-pane">
               <textarea id="signed-koblitz" class="codemirror-output"></textarea>
             </div>
+            -->
+            <div id="pane-signatures" class="tab-pane">
+              <div class="alert alert-info">
+                <strong>NOTE</strong>:
+                As of 2022-08-25, the former RSA and Bitcoin signature demos are
```
TallTed, line 568:
```suggestion
                As of 2022-08-25, the demos of RSA and Bitcoin signatures are
```

PR #792, opened 2022-09-02
Minor typo

nissimsan:


spec/latest/json-ld-framing/index.html
```diff
@@ -291,7 +291,7 @@ <h3>Framing</h3>
       using an example <a>frame</a> document which is used to both match the
       <a data-cite="JSON-LD11CG-API#dfn-flattened">flattened</a>
       data and show an example of how the resulting data should be shaped.
-      Matching is performed by using <a>properties</a> present in in the <a>frame</a>
+      Matching is performed by using <a>properties</a> present in the <a>frame</a>
       to find objects in the data that share common values. Matching can be done
       either using all properties present in the frame, or any property in the frame.
```
TallTed, line 296: might as well fix two at once...
```suggestion
      using either all properties present in the frame, or any property in the frame.
```

gkellogg:
Note that this is an old CG version of the Framing spec. If you go through the UI, it will direct you to the published location of the recommendation: https://www.w3.org/TR/json-ld-framing/, which has it's own GitHub repository.

To preserve history, the old CG versions have been left in place, but you shouldn't be able to navigate to them through the json-ld.org site navigation.

It might be a good idea to either add redirection headers to these pages, or do something like W3C does for obsolete specifications such as https://www.w3.org/TR/2020/PR-json-ld11-framing-20200507/, which we'd need to do manually, I suppose.

gkellogg:
See #793.

gkellogg:
Never mind, already done!

PR #794, opened 2022-09-10
Slides for TPAC Face to Face.

gkellogg:


TallTed:
This seems like a bunch of export artifacts, not anything editable/reviewable.  Perhaps most of these should be `.gitignore`d, and a few (or even, one) of them should be what we're seeing as a check-in?

gkellogg:
It’s the result of exporting the Keynote presentation to HTML, which I’ve done with previous presentation uploads. Presumably, these “artifacts” are used when viewing the slides. Ignoring some of them could lead to issues.

I’m not expecting a review if the details of the presentation, but of course anything in the presented slides can be fixed through a subsequent export.

PR #800, opened 2022-10-06
Add back forwarding rules for latest specs to /TR.

gkellogg:
This will forward (301) requests for the latest versions of specs to their /TR locations. Note that the site menu drop-downs already do this, as well as forward the latest drafts to w3c.github.io/...

I didn't actually remove the latest versions of the specs from the repo, but they shouldn't be accessible via HTTP. Neither did I attempt to add any robots.txt to keep them from being indexed.


msporny (approved):

iherman (approved):
Side issue: as soon as the RCH WG begins to produce readable documents, I guess the normalization and canonicalization URLs must be redirected, too...

pchampin (approved):

PR #808, opened 2023-07-01
Add two versions of Convenience Context

anatoly-scherbakov:
## Rationale

In YAML-LD, the `@` character is reserved. It is inconvenient to always quote JSON-LD `@`-keywords.

## What is done here

Depending on the domain, `@`-keywords can be mapped to various strings that would be more convenient for the writer. What we do here is propose two versions of such Convenience Contexts. They're usable both for JSON-LD and fr YAML-LD.


gkellogg (approved):

anatoly-scherbakov:
Renamed files as discussed, and also found that I had put them in the wrong place. Shall I merge?

@gkellogg

PR #811, opened 2023-07-31
Rename 'Github' to 'GitHub'

veyndan:
GitHub should be stylized with a capital "H".

Generated the PR by executed `find . -type f -exec sed -i '' "s/Github/GitHub/g" {} +` from the project root.

### DIFF
--- a/index.html
+++ b/index.html
@@ -409,7 +409,7 @@
         It is a <a href="http://www.w3.org/TR/json-ld/">W3C Recommendation</a>
         as of 16 January 2014. Participation is open
         to the public. There is a <a href="https://github.com/json-ld">JSON-LD
-        Github repository</a>. If you need immediate help, we have a <a
+        GitHub repository</a>. If you need immediate help, we have a <a
         href="https://irc.w3.org/?channels=#json-ld">#json-ld</a> IRC
         support channel on irc.w3.org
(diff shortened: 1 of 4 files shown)


davidlehn:
- I don't think any of the historical dated specs should be changed, regardless of the correction.  They are meant to be a snapshot in time.
- This adds an empty .ico file.

veyndan:
@davidlehn Updated the PR with the proposed changes

gkellogg (approved):
Other than index.html, the other files obsolete. Best practices should probably reference the WG Note.

davidlehn (approved):

veyndan:
> Other than index.html, the other files obsolete. Best practices should probably reference the WG Note.

@gkellogg How could I go about doing that? Would I just delete the non-`index.html` files here and replace the links with a reference to the WG note? What/where is the WG note?

gkellogg:
In other cases, we've both updated .htaccess to forward, and added some HTML. For example see https://github.com/json-ld/json-ld.org/blob/main/spec/latest/.htaccess.
Also, the test suite has an HTML redirect:

https://github.com/json-ld/json-ld.org/blob/f4b8dc9bb04ea3983a8c49bbe69913f6ff118989/test-suite/index.html#L10

PR #814, opened 2023-08-06
Delete primer

veyndan:
As recommended by https://github.com/json-ld/json-ld.org/pull/810#issuecomment-1666602859.


veyndan:
@gkellogg This deletion of primer was per your recommendation in https://github.com/json-ld/json-ld.org/pull/810#issuecomment-1666602859. Is there anything further that needs to be done in order to get this merged?

gkellogg (changes requested):
Actually, I think the group decided to obsolete the Best Practices document, not the primer, although neither is directly linked from json-ld.org. Once the group publishes https://w3c.github.io/json-ld-bp/ then https://json-ld.org/spec/latest/json-ld-api-best-practices/ can be removed. There is likely some content that needs to be moved over.

The primer at https://json-ld.org/primer/latest/ probably needs some TLC and should be better linked from json-ld.org.

gkellogg:
Looking at https://github.com/json-ld/json-ld.org/pull/810#issuecomment-1666602859 I suggested that the primer was obsolete, but I'm not sure we decided to just delete it.

BigBlueHat:
Closing this one as the action needs to be to obsolete the primer, not remove it.

PR #815, opened 2023-08-06
Remove reference to deleted `prettify.css`

veyndan:
`common/prettify.css` was deleted in 0702a1819008995a1052a675ea526d477c01feac, but some of those references remained.


BigBlueHat:
Thanks for finding that issue! It looks like @davidlehn commented those lines out in f98f5d9 ([for example](https://github.com/json-ld/json-ld.org/commit/f98f5d957a846e54da1c745fa29266eab70f4f50#diff-c30dfa14bbdbbaca7562a9dc460f275b98175bb26bc959bc5a3df823da7682dfL22-R22)) and I'll be removing them altogether soon. Appreciate the PR!

PR #821, opened 2023-09-11
Change CHAPI Playground links.

BigBlueHat:
The CHAPI Playground has moved to be the Verifiable Credentials Playground at https://vcplayground.org/


BigBlueHat:
Wow...I need more water and less ☕ We just did this via https://github.com/json-ld/json-ld.org/pull/820

PR #830, opened 2023-12-08
Add Sophia to list of implementations

pchampin:


gkellogg (approved):

index.html
```diff
@@ -367,6 +367,13 @@ <h1 class="span12" style="text-align: center;"> <span class="icon-cog"></span> D
                 </a>
                 <span property="schema:programmingLanguage" content="Rust"></span>
               </li>
+              <li typeof="schema:SoftwareSourceCode">
+                <a property="schema:codeRepository" href="https://crates.io/crates/sophia/0.8.0-alpha.3">
```
davidlehn, line 371: Would it be better to link to the main URL?  A version link will become stale.
https://crates.io/crates/sophia/

pchampin: The main URL currently point to version 0.7, which has very crude JSON-LD support compared to the (still alpha) 0.8. That's why I decided to put the explicit link to 0.8 instead.

I'l make an additional PR using the main URL once it points to 0.8 (after I release it, hopefully soon)

PR #836, opened 2024-03-19
Convert to Eleventy.

davidlehn:
Status:
- Second try at converting to a static site generator.
  - First try was to Jekyll in 2017: https://github.com/json-ld/json-ld.org/pull/461.
  - Much of the discussion there applies here after many years.
- There are many many many SSG choices. This conversion is using [Eleventy](https://www.11ty.dev/). If the community wants to use another system, it's likely these conversions are a good start with similar patterns.
- What's happening in the initial commits here:
  - Handles pass through of all useful (and more) files that were served before.
  - Converts the runtime php indexing code to build time js code.
  - Some tweaks for uniformity.
- The plain apache+php hosting we use now could probably host this output with a few operational changes.
- Templates and various upgrades and improvements are not yet done.  Solving issues one step at a time.

Issues:
- Basically same ones as in the jekyll issue.
- How to handle `minutes/`.  It's gigantic and inappropriate to mix into a static site directly.
  - It is being updated on the main site AND on gh pages at https://json-ld.github.io/minutes/, but the logs note that 1.8G is beyond the 1G max and deploy might fail and it can take 20min+ to deploy.  Another solution would be better.  Also the "pre-wg" link is easy to not notice.
- How to handle the playground and playground php proxies.
- How best to host and handle updates.
  - May need content type and cors fixes depending on hosting platform.

Notes:
- One approach for many issues is a subdomain using gh pages or something else for minutes, cloudflare for hosting, cloudflare workers for the proxy.  I think Benjamin likes this idea.
- We also need to decide how to improve the playground.  No one at all wants to work on it as is.  Integrating a web app build step into the static site process is doable but perhaps not the best approach.  I'm pondering the site code here being more content based and (ab)using a CDN or gh pages to host built playground app bundles and resources.  Suggestions welcome on that.

Testing:
```sh
npm install
npm run serve
# open http://localhost:8080/ (or whatever URL it gives you)
```


gkellogg (approved):
Thanks for work on this, Dave! Most of the space the minutes consume is due to audio captures of the early meetings, which we haven’t kept going in some time. Do logs indicate if they’re ever actually accessed? Maybe there’s a way to move those someplace else off of the main (or minutes) site.

.eleventy.js
```diff
@@ -0,0 +1,52 @@
...
+  eleventyConfig.ignores.add('contexts/person.html');
+  eleventyConfig.ignores.add('examples');
+  eleventyConfig.ignores.add('images/README.md');
+  eleventyConfig.ignores.add('minutes/**/*');
+  eleventyConfig.ignores.add('ns/json-ld.html');
+  eleventyConfig.ignores.add('playground/dev/README.md');
+  eleventyConfig.ignores.add('presentations');
+  eleventyConfig.ignores.add('scripts');
+  eleventyConfig.ignores.add('spec/tools');
+  eleventyConfig.ignores.add('spec/LICENSE.md');
```
BigBlueHat, line 47: These ignores can be moved to `.eleventyignore` to make this file more sane.

package.json
```diff
@@ -0,0 +1,15 @@
...
+  "description": "json-ld.org homepage",
+  "scripts": {
+    "build": "npx @11ty/eleventy",
+    "serve": "npx @11ty/eleventy --serve",
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "",
+  "devDependencies": {
+    "@11ty/eleventy": "^2.0.1"
```
BigBlueHat, line 13: Any reason to use both `npx` and declare the dependency?

package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "json-ld.org",
+  "private": true,
+  "description": "json-ld.org homepage",
+  "scripts": {
+    "build": "npx @11ty/eleventy",
+    "serve": "npx @11ty/eleventy --serve",
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
```
BigBlueHat, line 10: This should probably be...
```suggestion
  "author": "Digital Bazaar, Inc.",
```
...or if we consider this a donation, then "W3C, Inc."

package.json
```diff
@@ -0,0 +1,15 @@
...
+  "license": "",
```
BigBlueHat, line 11: We should switch this repo to the `BSD-3-Clause`. The W3C software and documentation license has increasingly fallen out of favor in place of ye olde (and far more common) BSD (thankfully).

BigBlueHat (approved):
Looks great overall. Just some non-blocking recommendations.

package.json
davidlehn: Removing the unused field.  This project uses package.json for build rules, not metadata.

package.json
davidlehn: Removing the unused field.  This project uses package.json for build rules, not metadata.  There exists a LICENSE.md file for license details.

package.json
davidlehn: No reason.  Switching to remove npx call.

.eleventy.js
davidlehn: What's wrong with having them here?  At least for now.  It's a bit confusing as is due to eleventy processing various things, having to pass through others, and ignore particular files.  And for drafts, that's done computationally.  I thought it made more sense to keep the details in one place.

davidlehn:
Another issue with this conversion is how to handle all the `.htaccess` rules.  If the target host is Apache then of course these will still work.  But our likely new target is Cloudflare Pages, which does things differently.

```
$ git ls-files | grep .htaccess
.htaccess
contexts/.htaccess
images/.htaccess
spec/ED/.htaccess
spec/latest/.htaccess
test-suite/.htaccess
test-suite/tests/.htaccess
utils/.htaccess
```

- CF has `_redirects` and `_headers` files, which are simpler and limited.  Some of the `.htaccess` rules can be handled with those files.  (Untested, but looks promising)
- The content type detection is trickier and I believe would require another `_functions` worker for some special cases.  Effectively writing some js to do what apache rules do.

And while pondering these things, I'm wondering where this is headed.
- A good dev env is now rather complex.
- 11ty has the nice auto update dev server.  But it's unaware of these htaccess or cloudflare rules.  Making it difficult to actually test the entire site.  To use the playground proxy (see other PR), you need to have a 11th proxy to a cloudflare wrangler dev server.
- Cloudflare wrangler could also serve the static dev site with its redirects and headers and functions.  But then the auto update feature doesn't work.  There may be some extra proxying to make that work.  But then that would require two dev servers to do dev work, which seems a bit awkward.
- Should we make apache or cloudflare a build option?  I suspect it will just flip which files get copied to the output.  But that's two things to maintain and keep in sync.

.eleventy.js
BigBlueHat: I get the single place to look. I mostly just like data in inert places whenever possible vs. buried among API calls.

BigBlueHat:
@davidlehn per your concerns above, I think we can keep using 11ty separate from the hosting platform. Having 11ty in place makes working on (and redesigning) the site much simpler--where it's hosted can be a separate matter because 11ty generates static sites.

Cloudflare hosting can easily get more complex/tangled (just like the build tool nightmare that is modern JavaScript vs. ye olde `<script>` tag). It really depends on what we want to manage, pay for, support, etc. Most/all of the Cloudflare stuff, though, can also be defined descriptively (or at least via JS in the same repo) to avoid "blind"/hidden configurations available only via the Cloudflare Dashboard.

All that to say, how this is hosted is orthogonal to using 11ty.

README.md
```diff
@@ -49,15 +54,15 @@ The Specifications
 If you are a developer, you may be interested in the official JSON-LD W3C
 specifications:
 
-* `JSON-LD 1.1 - A JSON-based Serialization for Linked Data`_
-* `JSON-LD 1.1 Processing Algorithms and API`_
-* `JSON-LD 1.1 Framing`_
+* [JSON-LD 1.1 - A JSON-based Serialization for Linked Data][]
+* [JSON-LD 1.1 Processing Algorithms and API][]
+* [JSON-LD 1.1 Framing][]
```
TallTed, line 59:
```suggestion
* [JSON-LD 1.1 - A JSON-based Serialization for Linked Data](http://www.w3.org/TR/json-ld/)
* [JSON-LD 1.1 Processing Algorithms and API](https://www.w3.org/TR/json-ld-api/)
* [JSON-LD 1.1 Framing](https://www.w3.org/TR/json-ld-framing/)
```

README.md
```diff
@@ -49,15 +54,15 @@ The Specifications
...
 https://json-ld.org/spec/
 
-.. _list of users: https://github.com/json-ld/json-ld.org/wiki/Users-of-JSON-LD
-.. _JSON-LD 1.1 - A JSON-based Serialization for Linked Data: http://www.w3.org/TR/json-ld/
-.. _JSON-LD 1.1 Processing Algorithms and API: http://www.w3.org/TR/json-ld-api/
-.. _JSON-LD 1.1 Framing: http://www.w3.org/TR/json-ld-framing/
+[list of users]: https://github.com/json-ld/json-ld.org/wiki/Users-of-JSON-LD
+[JSON-LD 1.1 - A JSON-based Serialization for Linked Data]: http://www.w3.org/TR/json-ld/
+[JSON-LD 1.1 Processing Algorithms and API]: https://www.w3.org/TR/json-ld-api/
+[JSON-LD 1.1 Framing]: https://www.w3.org/TR/json-ld-framing/
```
TallTed, line 68: What's intended here? Human text followed by visible URL, or human text anchored with an `href`? If the former, remove the `[]` that now surrounds the human text; if the latter, remove the `: ` and wrap the URL in `(). I'm guessing the latter is the intent, so --
```suggestion
[list of users](https://github.com/json-ld/json-ld.org/wiki/Users-of-JSON-LD)
[JSON-LD 1.1 - A JSON-based Serialization for Linked Data](http://www.w3.org/TR/json-ld/)
[JSON-LD 1.1 Processing Algorithms and API](https://www.w3.org/TR/json-ld-api/)
[JSON-LD 1.1 Framing](https://www.w3.org/TR/json-ld-framing/)
```

TallTed: But if it's the former --
```suggestion
**list of users:** https://github.com/json-ld/json-ld.org/wiki/Users-of-JSON-LD
**JSON-LD 1.1 - A JSON-based Serialization for Linked Data:** http://www.w3.org/TR/json-ld/
**JSON-LD 1.1 Processing Algorithms and API:** https://www.w3.org/TR/json-ld-api/
**JSON-LD 1.1 Framing:** https://www.w3.org/TR/json-ld-framing/
```

TallTed (changes requested):

BigBlueHat:
@davidlehn how close do you think this is to shippable? I know #840 is part of the consideration here too. Just curious how we can get this over the finish line.

Since the WG is getting rechartered, it'd be super to get this in a space we can continue to improve, redesign, etc.

Happy to help finish it off...just not sure where we're stuck.

README.md
BigBlueHat: Same situation as below: https://github.com/json-ld/json-ld.org/pull/836#discussion_r2016923384

README.md
TallTed: Oh, got it. More "Markdown making it not-really-simpler".

README.md
TallTed: Acked.

BigBlueHat:
> * 11ty has the nice auto update dev server.  But it's unaware of these htaccess or cloudflare rules.  Making it difficult to actually test the entire site.  To use the playground proxy (see other PR), you need to have a 11th proxy to a cloudflare wrangler dev server.
> * Cloudflare wrangler could also serve the static dev site with its redirects and headers and functions.  But then the auto update feature doesn't work.  There may be some extra proxying to make that work.  But then that would require two dev servers to do dev work, which seems a bit awkward.

To untangle this, I've added a `npm run pages` option which can be used alongside `npm run watch`. That should provide the handy auto-building from 11ty while serving a localhost Cloudflare-like environment.

> * Should we make apache or cloudflare a build option?  I suspect it will just flip which files get copied to the output.  But that's two things to maintain and keep in sync.

I don't think we should do this as it'd be a good deal of overhead to maintain/juggle both. Consequently, I've removed the `.htaccess` files in #855.

Once #855 is merged into this PR, I think this one can be marked "ready for review", and we can get a preview URL on Cloudflare sporting all the things for further testing from everyone here.

TallTed (approved):
My earlier concerns were due to misinterpreting Markdown. They're resolved

BigBlueHat:
@davidlehn I think I've addressed most if not all of your concerns in https://github.com/json-ld/json-ld.org/pull/836#issuecomment-2103040505 via #855, so I've marked this as ready for review.

Please let me know what else it needs. Thanks!

BigBlueHat:
Clicking back through the site, I realized we never answered this one:
> How to handle `minutes/`. It's gigantic and inappropriate to mix into a static site directly.
> * It is being updated on the main site AND on gh pages at https://json-ld.github.io/minutes/, but the logs note that 1.8G is beyond the 1G max and deploy might fail and it can take 20min+ to deploy.  Another solution would be better.  Also the "pre-wg" link is easy to not notice.

https://99ef1971.json-ld-org.pages.dev/minutes/ (in the preview) is current 404.

The simplest solution would be to setup a redirect from that URL to https://json-ld.github.io/minutes/ (where it's actually hosted).

Alternatively, if we think the json-ld.org URLs are important, we could proxy those URLs through a Cloudflare Pages Function (though that likely comes with some bandwidth costs)--which should not incur additional costs (afaict from https://developers.cloudflare.com/workers/platform/pricing/).

Or, similarly, we could push the files to a Cloudflare R2 public bucket (which supports large static file hosting similar to S3). However, that _may_ eventually incur some cost or at minimum we'd have to be strategic in how we setup publishing the files to R2--so we don't ship all 1.8 GB every time a new change is made to the minutes...

My current vote would be a simple redirect setup to keep thing uncomplicated.

Thoughts @davidlehn @gkellogg?

BigBlueHat:
It's time.

PR #840, opened 2024-05-01
Convert proxy to Cloudflare Pages Function.

davidlehn:
- Development of the proxy is a bit awkward.  Need to run two servers and let the Eleventy dev server proxy to the real proxy on a wrangler dev server.  Only needed if testing the `http:` support locally. The people doing that will probably rarely be more than me.
- The Eleventy proxy bits require a newer feature in the Eleventy 3.x dev server.  The rest of the site seems to build with that.
- Suggestions welcome for hardening the proxy to be not useful outside of the playground.


.eleventy.js
```diff
@@ -49,4 +53,39 @@ module.exports = function(eleventyConfig) {
...
+    eleventyConfig.setServerOptions({
+      onRequest: {
+        '/playground/proxy': playgroundProxy
+      }
+    });
+  }
 };
+
+// proxy to worker proxy
+async function playgroundProxy({url}) {
```
BigBlueHat, line 68: Can this be defined above its use on line 61?

README.md
```diff
@@ -62,7 +62,37 @@ A list of all previous specification drafts is also available.
...
+
+[Cloudflare Pages Functions]: https://developers.cloudflare.com/pages/functions/
+[Cloudflare Pages]: https://pages.cloudflare.com/
+[Eleventy]: https://www.11ty.dev/
 [JSON-LD 1.1 - A JSON-based Serialization for Linked Data]: http://www.w3.org/TR/json-ld/
-[JSON-LD 1.1 Processing Algorithms and API]: https://www.w3.org/TR/json-ld-api/
 [JSON-LD 1.1 Framing]: https://www.w3.org/TR/json-ld-framing/
+[JSON-LD 1.1 Processing Algorithms and API]: https://www.w3.org/TR/json-ld-api/
+[Wrangler]: https://developers.cloudflare.com/workers/wrangler/
+[list of users]: https://github.com/json-ld/json-ld.org/wiki/Users-of-JSON-LD
```
BigBlueHat, line 97: Where did this sneak in from? Just a missing link?

functions/playground/proxy.js
```diff
@@ -0,0 +1,68 @@
+// json-ld.org playground proxy
+//
+// details:
+// - Basic proxy for use with the json-ld.org playground.
+// - Built for deployment using Cloudflare Pages Functions API
+//   - https://developers.cloudflare.com/pages/functions/
+// - Only handle GET requests.
+// - Only designed to be used by the playground, no CORS support.
+// - Only handle requests for 'http:' URLs.
```
BigBlueHat, line 9: Can you explain the `http:` only thing? I'm confused by how this'll work in practice since context URLs are often served from `https:` URLs (and should be). I'm likely missing something...

BigBlueHat (changes requested):
One code order tweak, and a couple questions. Thanks for continuing to push this forward!

README.md
davidlehn: Missing?  It's used on line 21.

functions/playground/proxy.js
davidlehn: The purpose of the proxy is to avoid browser security errors and allow the https site to load http resources.  If you look at playground.js you can see that's the only time it's used.  Best practice these days is to use https, but that's not required, and historically it was enough of an issue that we wrote the php proxy.

BigBlueHat (approved):

BigBlueHat:
Works great locally. I'm merging this into `11ty` and we can continue to improve it there if/as needed.

package.json
```diff
@@ -5,9 +5,11 @@
   "scripts": {
     "build": "eleventy",
     "serve": "eleventy --serve",
+    "watch": "eleventy --watch",
+    "dev": "wrangler pages dev _site --compatibility-date=2024-04-27 --live-reload --port 8788",
```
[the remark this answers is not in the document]
davidlehn: Do consider that wrangler is not a small dependency.  Might be better to use npx or just tell people to globally (or otherwise) install it themselves.

PR #851, opened 2024-12-06
convert `remoteDoc.document` if string

caspervk:


caspervk:
Updated PR description. Apologises for not properly describing the issue and attempted solution sooner.

PR #852, opened 2024-12-14
Add missing purpose field to t0041

multimeric:
Test `t0041` had two `name` fields and no `purpose` field, which was causing my code to break. I assume it was a mistake. I've fixed that here.


davidlehn (approved):

gkellogg:
Tests live at https://w3c.github.io/json—ld-api/tests. The ones in this repo are archaic. The tests in this repo should be removed. The test there does not have a duplicate name, but is missing a purpose field.

multimeric:
Can you point me to the repo where the tests live please? I found https://github.com/json-ld/tests but that's also obsolete.

multimeric:
I suppose it's here? https://github.com/w3c/json-ld-api/tree/main/tests

PR #859, opened 2025-04-23
Add redir from /minutes/ to minutes subdomain.

BigBlueHat:


BigBlueHat:
You can test a redirect with https://add-minutes-redirect.json-ld-org.pages.dev/minutes/2022-09-12/

dlongley (approved):

davidlehn (approved):
The _redirects ordering is odd.

BigBlueHat:
> The _redirects ordering is odd.

We can resort them, but anything dynamic should go last as it's a first-come-first-served thing and dynamic stuff takes longer (as anywhere) than static matches.

The order they're in now is based on the order I handled the .htaccess migration...no more no less (other than the above caveat).

PR #862, opened 2025-05-05
Setup 301 redirect for /test-suite/.

BigBlueHat:
The current page uses a meta refresh (after 5 seconds) to go to the API test suite. Consequently, this page serves little purpose now, so converting it to a full 301 redirect seems to make the most sense.

Additionally, I'm not sure we need the `test-suite/` directory here at all sense it seems all test suite work is no in each spec's specific repo.

Thoughts?


gkellogg (approved):

BigBlueHat:
@gkellogg do you think we can/should remove the `test-suite/` directory altogether? Do we use it for anything? It seems to have all been moved elsewhere.

PR #872, opened 2025-05-08
Playground Next

BigBlueHat:
~Still early days, but definitely a more stable foundation.~

Completely rebuilt Playground at last!

Fomantic-UI is used for the design of this page (to be used elsewhere once the
new playground is deployed--in a separate PR).

The editor is now [CodeMirror](https://codemirror.net/) 6 and the UX integration code is using [petite-vue](https://github.com/vuejs/petite-vue).

You can preview the new Playground at https://playground-next.json-ld-org.pages.dev/playground/next/


BigBlueHat:
Thanks, @gkellogg .I'll watch for that in this new Playground.

BigBlueHat:
The only remaining (functional) tab I need to reimplement is the "Visualized" tab. I plan to move the notice for "Signatures" now being on the VC Playground into the notice box at the top in order to gain more horizontal tab area.

So, the only features left to implement (afaict) are:
- the Visualized tab
- remote document / context loading

Features I do _not_ plan to implement (yet...at least) are URL shortening and Gist creation. Both are broken in various ways on the current site, and reimplementing them seems like a much lower priority than adding YAML-LD and CBOR-LD output tabs (and eventually input options)--both of which I want to address after this becomes the main playground.

Consequently, @gkellogg @davidlehn @anatoly-scherbakov @niklasl @TallTed, please test this thoroughly! ❤️

BigBlueHat:
New features include a YAML-LD output tab (hi there, @anatoly-scherbakov) and a CBOR-LD output tab!

I'm planning to do the Visualized tab later as the visualization library hasn't been updated in ages, so I'm expecting that to be a relatively heavy lift...so I don't want to block on that, but plan to return to it soon (as I find it useful).

So, with that status, this thing feels ready to go and I'm itching to get it in front of developers sooner than later--so reviews welcome!

BigBlueHat:
...actually...looks like I introduced a bug into the editing when I added the debounce stuff... Trying to sort that out.

gkellogg:
> k, folks. I need to stop making this more awesome, so y'all can review it! 😜

It is, indeed, looking pretty awesome!

> This should now have greatly improved autocomplete as well as syntax highlighting!
> 
> Other than fixing bugs here, I'm keen to get this merged and do any more feature addition/improvements on other branches...so, please review soon! 😁

It would be nice if the input could be YAML-LD or CBOR-LD, as well as N-Quads. Of course, even nicer, if input could be any RDF format (and output), at least TriG.

gkellogg (approved):

BigBlueHat:
> It would be nice if the input could be YAML-LD or CBOR-LD, as well as N-Quads. Of course, even nicer, if input could be any RDF format (and output), at least TriG, using for example [RDF JS N3.js](https://github.com/rdfjs/N3.js).

Want to make some issues? 😁

gkellogg:
> > It would be nice if the input could be YAML-LD or CBOR-LD, as well as N-Quads. Of course, even nicer, if input could be any RDF format (and output), at least TriG, using for example [RDF JS N3.js](https://github.com/rdfjs/N3.js).
> 
> Want to make some issues? 😁

See https://github.com/json-ld/json-ld.org/issues/875.

Issue #875, opened 2025-06-04
Alternative formats for playground inputs


gkellogg:
The playground has only allowed JSON-LD to be used as an input, or a context. Given other formats (YAML-LD and CBOR-LD), it would be useful to be able to specify the input as an alternative format. For CBOR-LD, this should probably use an ASCII visualization of the binary form (Annotated HEX and/or Diagnostic Notation).

The input should also allow an RDF format as input. Nominally, N-Quads, but TriG would also be useful. It would also be nice to be able to view the output as TriG. when appropriate.

PR #881, opened 2025-11-04
Parse YAML workaround. See #875.

ioggstream:
## This PR

- modify the parser to accept YAML instead of JSON

## Notes

I tried to bundle the editor with npm run build:editor, but it does not seem to work on my PC, while just modifying the .js works.

### DIFF
--- a/playground/next/editor.mjs
+++ b/playground/next/editor.mjs
@@ -98,7 +98,7 @@ function editorListener(docName) {
           return;
         } else {
           try {
-            const parsed = JSON.parse(latestChange);
+            const parsed = YAML.parse(latestChange);
             this[docName] = parsed;
             this.parseError = {};
             this.setOutputTab(this.outputTab);


anatoly-scherbakov (approved):

stefanone91:
> @stefanone91 thanks for the additional workflow, but we're currently publishing this via Cloudflare Pages due to needing some additional HTTP headers on certain resources: https://github.com/json-ld/json-ld.org/blob/main/_headers
> 
> Would you mind removing the workflow? Or at least putting it in a draft PR, so we can discuss that separately?

@BigBlueHat yes sure! I'm sorry. I'll split in a separate PR.

ioggstream:
Thx @stefanone91 :)

BigBlueHat:
So, I did a good bit of digging on this today and...there's a great deal more to be done to wire up CodeMirror properly so that it doesn't throw JSON errors in the YAML.

Additionally, if we're really aiming at #875 then we'll need to either provide UX for switching formats (my original plan) or do some sort of file detection (which seems...fraught). Likely a mix of both could work...maybe.

For now, I'm closing this PR to make room for others. Feel free to continue the conversation on #875.

Cheers!

ioggstream:
We are experimenting on our fork here https://github.com/ioggstream/json-ld.org
that is YAML only for now.

When we'll have something decent, we'll PR again on this topic :)

Have a nice day,
R.

cc: @stefanone91

PR #883, opened 2025-11-18
update all the things

BigBlueHat:


davidlehn:
Why is the built `editor.bundle.js` file in git?  I didn't notice that before.  Looks like it could be built into the output _site dir and avoid huge churn in git history.

PR #889, opened 2025-12-04
Update example Lennon spouse

### LINKED ISSUE #880: Lennon - spouse (Yoko Ono) - example
Noticing that the example of linked data prominent on the front page of the site and the README contains an error that might show implicit bias:

Image

There is no reason that a singular "spouse" link for John Lennon would be a link to Cynthia_Lennon (first spouse, divorced at time of death) as opposed to Yoko_Ono (spouse at death).  

It is important that the contributions of Asians and Asian Americans to history (including music history) not be minimized but the current state of the example page does just that.  I'm sure it was unintentional but hope it can be fixed.  Thanks!

mscuthbert:
See discussion at #880  (Fixes #880)

Only changed the index page, README, example 91, and spec latest -- left historical examples alone.

Thanks!

### DIFF
--- a/examples/syntax/example-091-Embedding-JSON-LD-in-HTML.html
+++ b/examples/syntax/example-091-Embedding-JSON-LD-in-HTML.html
@@ -4,6 +4,7 @@
     "@id": "http://dbpedia.org/resource/John_Lennon",
     "name": "John Lennon",
     "born": "1940-10-09",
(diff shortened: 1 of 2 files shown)


BigBlueHat:
@mscuthbert check-out #894. I'm basically just avoiding the problem altogether. 😅

PR #890, opened 2025-12-04
Add markdown-it plugins.

davidlehn:
- Use `markdown-it-anchor`.
- Use `markdown-it-attrs`.

Used by upcoming PRs.


BigBlueHat (approved):
Nice find!

PR #894, opened 2025-12-05
update homepage example

### LINKED ISSUE #880: Lennon - spouse (Yoko Ono) - example

BigBlueHat:
- **Describe the JSON-LD homepage in the main example.**

Fixes #880 by avoiding the problem altogether.

### DIFF
--- a/index.liquid
+++ b/index.liquid
@@ -21,7 +21,11 @@ masthead:
     </p>
   </div>
   <div class="eight wide column">
-    <pre class="ui segment">{
+    <div class="ui segment">
+      <a class="ui top right attached primary label"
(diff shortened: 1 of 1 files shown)


mscuthbert:
It does fix the problem from #889 - but I don't know if this is the best direction for the project.  With the Lennon example I knew exactly something I could do with json-ld -- "ah, it's a way of representing metadata and linked data about people...and probably other things...in JSON! Cool"

When I see a site that describes its own website...I wouldn't think that this is the proper library for ''my'' purposes --  it brings out a lot the **meta** in metadata.

Why not describe an animal instead?  Common-name, species, etc. Egg-laying:boolean?  As an added bonus it immediately becomes a symbol of the project!  (asked ChatGPT for common animals not already closely associated with particular tech projects and Otter, Swan, and Barn Owl came up ... along with more specific things like Tanuki (Racoon dog -- my fav), Agouti,Opah ("bonus: natural circle-logo shape"), etc.

davidlehn:
- That "now" thing isn't needed in the update.
- Does this make things better or worse?  Someone unfamiliar with intricacies of JSON-LD might ask if those `spouse` values are concurrent or ordered, and if ordered, why are they reversed.  All this can be explained, and may open a discussion about data modeling, but I'd think the main front page example should avoid any of that.
- I'll again suggest we use something without complexities.  If birthPlace was added to that context, it would be simple enough.

PR #901, opened 2025-12-16
fix options

BigBlueHat:
- **Fix base URL; simplify UX.**
- **Improve Safe Mode error display.**

### DIFF
--- a/playground/next/editor.mjs
+++ b/playground/next/editor.mjs
@@ -505,7 +505,7 @@ window.app = createApp({
         // TODO: this should happen elsewhere...like a watcher
         try {
           const output = await jsonld.canonize(this.doc, {
-            format: 'application/n-quads', ...this.options
+            format: 'application/n-quads', ...{...this.options, ...{safe: true}}
           });
           setEditorValue(readOnlyEditor, output);
           this.parseError = {};


BigBlueHat:
> * How is that going to work to let people see what default behavior is?

I guess there is some risk here if the defaults change since they're being "matched" in the UX code rather than just not passing them in. Would calling out which one is the default help? Also, are these defaults unique to `jsonld.js`?

I have a longer term goal (now) to incorporate other libraries ([Sophia](https://github.com/pchampin/sophia_rs) via [SoWASM](https://github.com/pchampin/sowasm?tab=readme-ov-file) and [ldtr](https://github.com/niklasl/ldtr) at least).

I want to be sure what the UX's defaults are clear...more than I care what the underlying libraries defaults are...if that makes sense.

BigBlueHat:
> I did hope some of the playground updates could mostly match previous behavior as a first update step. Then work on these sorts of design decisions next.

That's been my general intent. I changed course a bit on this one because it was more work (which I would shortly remove...) to rebuild the more confusing 3 choices for 2 options UX.

@davidlehn perhaps as a compromise I could mark which of the options are the default?

I agree some sort of "Safe mode enabled" notice/banner/thing would be useful to state on the `Canonized` tab. Additionally, is there any reason to use `canonize` _without_ safe mode being enabled? Because the current code (which matches previous behavior) currently prevents it ever being toggled off.

PR #905, opened 2026-01-23
Set the `@base` to current URL.

BigBlueHat:
This is what an empty value technically does, but setting it in the form data makes that clearer than hiding it in the guts of the machine.

It can be overridden via that same form field.


BigBlueHat:
> I'm just looking at the source, mostly for English. No _running_/compliation of that source.

It's available at https://set-base-to-current-url.json-ld-org.pages.dev/playground/next/ fwiw. No need to download it other than into your browser like any other webpage. 😉

PR #907, opened 2026-01-26
Remove auto-setting compaction context.

BigBlueHat:
There are good reasons to leave it as `{}` and this code prevented that value ever being possible, so removing it.

This PR can be tested at https://fix-compaction-context.json-ld-org.pages.dev/playground/next/


azaroth42 (approved):
Agreed remove. 

Future feature might make this a setting somewhere, but currently being unable to test compaction with an empty context seems wrong.

PR #917, opened 2026-02-23
Bring back Visualized tab.

BigBlueHat:
Brought back the "Visualized" tab by resurrecting [jsonld-vis](/json-ld/jsonld-vis) (forked into this org if anyone wants to help 😁).


BigBlueHat:
> * Isn't rolldown the new bling?

Maybe. Not sure it's worth the lift to do it again, though.

